PR #27, opened 2022-09-16
Support API Rules on the same path and different methods

cnvergence:
**Description**

- Add check for path duplication
- Validate same path and different method
- Filter out path duplication for virtual services

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/6830


werdes72 (approved):
/woof

Issue #34, opened 2022-09-22
[MVP] JWT Istio handler configuration - support for new APIRules 


strekm:
**Description**

Implement changes in jwt access strategy handler based on Istio CRs. jwt access strategy should support both methods of handling: by creating ORY oathkeeper rule and by creating Istio RA and AP CRs. This version should support only creation of APIRule, transition between handlers is not supported. First version should be implemented as follows:

```
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: service-secured
spec:
  gateway: kyma-system/kyma-gateway
  host: foo.bar
  service:
    name: foo-service
    port: 8080
  rules:
    - path: /headers
      methods: ["GET"]
      mutators: []
      accessStrategies:
        - handler: jwt
          config:
            authentications:
            - issuer: ISSUER
              jwksUri: JWKSURI
            - xxx
```
`config` object will contain:
- `authentications` - required array with at least one element

`authentication` object will contain:
- `issuer` - required field  (this field is not required for oathkeeper jwt handler)
- `jwksUri` - required field (this field is required for oathkeeper jwt handler)

This configuration should create following resources:
```
apiVersion: security.istio.io/v1beta1
kind: RequestAuthentication
metadata:
  name: httpbin
  namespace: foo
spec:
  selector:
    matchLabels:
      app: httpbin
  jwtRules:
  - issuer: ISSUER
    jwksUri: JWKSURI
```

This configuration should create following AP:
```
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: httpbin
  namespace: foo
spec:
  selector:
    matchLabels:
      app: httpbin
  rules:
  - from:
    - source:
        requestPrincipals: ["*"]
    to:
    - operation:
        methods: ["GET"]
        paths: ["/headers"]
```

If FT is disabled oathkeeper handler schema should be valid. If FT is enabled jwt handler schema should be valid.

ACs:
- [x] othkeeper implementation fully supported
- [x] istio implementation supported for defined fields
- [x] istio implementation only applied when feature toggle is enabled
- [x] jwt Istio based handler only supported for new APIRules (transition not supported)
- [x] jwt handler configuration is validated, errors are stored in Status

**Reasons**

Introducing JWT handler based on Istio.

**Attachments**

Depends on: https://github.com/kyma-project/api-gateway/issues/40
Part of: https://github.com/kyma-project/api-gateway/issues/30

[...]

PR #53, opened 2022-10-21
JWT Istio handler configuration

werdes72:
See: #34


internal/processing/processing.go
```diff
@@ -59,15 +60,25 @@ func (f *Factory) CalculateRequiredState(api *gatewayv1beta1.APIRule, config *he
...
+	res.authorizationPolicies = make(map[string]*istiosecurityv1beta1.AuthorizationPolicy)
+	res.requestAuthentications = make(map[string]*istiosecurityv1beta1.RequestAuthentication)
+
 	for _, rule := range api.Spec.Rules {
 		if isSecured(rule) {
 			var ar *rulev1alpha1.Rule
+			var ap *istiosecurityv1beta1.AuthorizationPolicy
+			var ra *istiosecurityv1beta1.RequestAuthentication
+
 			if config.JWTHandler == helpers.JWT_HANDLER_ORY {
```
barchw, line 72: It would be good to use some switch statement like:
```go
switch config.JWTHandler {
case helpers.JWT_HANDLER_ORY:
	ar = generateAccessRule(api, rule, rule.AccessStrategies, f.additionalLabels, f.defaultDomainName)
	res.accessRules[setAccessRuleKey(pathDuplicates, *ar)] = ar
case helpers.JWT_HANDLER_ISTIO:
	//TODO generate based on config.JWTHandler="istio"
	ap = generateAuthorizationPolicy(api, rule, f.additionalLabels)
	ra = generateRequestAuthentication(api, rule, f.additionalLabels)
	res.authorizationPolicies[setAuthorizationPolicyKey(pathDuplicates, ap)] = ap
	res.requestAuthentications[setRequestAuthenticationKey(ra)] = ra
}
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
+package builders
+
+import (
+	"istio.io/api/security/v1beta1"
+	apiv1beta1 "istio.io/api/type/v1beta1"
+	securityv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
+)
+
+// AuthorizationPolicyBuilder returns a builder for istio.io/apis/security/v1beta1/AuthorizationPolicy type
```
videlov, line 9:
```suggestion
// AuthorizationPolicyBuilder returns a builder for istio.io/client-go/pkg/apis/security/v1beta1/AuthorizationPolicy type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+	aps.value.Selector = val.Get()
+	return aps
+}
+
+func (aps *AuthorizationPolicySpec) Rule(val *Rule) *AuthorizationPolicySpec {
+	aps.value.Rules = append(aps.value.Rules, val.Get())
+	return aps
+}
+
+// RuleBuilder returns builder for istio.io/apis/security/v1beta1/Rule type
```
videlov, line 93:
```suggestion
// RuleBuilder returns builder for istio.io/api/security/v1beta1/Rule type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+	r.value.From = append(r.value.From, val.Get())
+	return r
+}
+
+func (r *Rule) RuleTo(val *RuleTo) *Rule {
+	r.value.To = append(r.value.To, val.Get())
+	return r
+}
+
+// RuleFromBuilder returns builder for istio.io/apis/security/v1beta1/Rule_From type
```
videlov, line 118:
```suggestion
// RuleFromBuilder returns builder for istio.io/api/security/v1beta1/Rule_From type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+func (rt *RuleTo) Get() *v1beta1.Rule_To {
+	return rt.value
+}
+
+func (rt *RuleTo) Operation(val *Operation) *RuleTo {
+	rt.value.Operation = val.Get()
+	return rt
+}
+
+// OperationBuilder returns builder for istio.io/apis/security/v1beta1/Operation type
```
videlov, line 160:
```suggestion
// OperationBuilder returns builder for istio.io/api/security/v1beta1/Operation type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+	o.value.Methods = val
+	return o
+}
+
+func (o *Operation) Path(val string) *Operation {
+	o.value.Paths = append(o.value.Paths, val)
+	return o
+}
+
+// RequestAuthenticationBuilder returns a builder for istio.io/apis/security/v1beta1/AuthorizationPolicy type
```
videlov, line 185:
```suggestion
// RequestAuthenticationBuilder returns a builder for istio.io/client-go/pkg/apis/security/v1beta1/AuthorizationPolicy type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+	ra.value.Labels[key] = val
+	return ra
+}
+
+func (ra *RequestAuthentication) Spec(val *RequestAuthenticationSpec) *RequestAuthentication {
+	ra.value.Spec = *val.Get()
+	return ra
+}
+
+// RequestAuthenticationSpecBuilder returns a builder for istio.io/api/type/v1beta1/RequestAuthentication type
```
videlov, line 239:
```suggestion
// RequestAuthenticationSpecBuilder returns a builder for istio.io/api/security/v1beta1/RequestAuthentication type
```

internal/builders/istio.go
```diff
@@ -0,0 +1,305 @@
...
+	ras.value.Selector = val.Get()
+	return ras
+}
+
+func (ras *RequestAuthenticationSpec) JwtRules(val *JwtRule) *RequestAuthenticationSpec {
+	ras.value.JwtRules = *val.Get()
+	return ras
+}
+
+// JwtRuleBuilder returns builder for istio.io/api/type/v1beta1/WorkloadSelector type
```
videlov, line 269:
```suggestion
// JwtRuleBuilder returns builder for istio.io/api/security/v1beta1/JWTRule type
```

internal/processing/authorization_policy_helpers.go
```diff
@@ -0,0 +1,45 @@
...
+	for k, v := range additionalLabels {
+		arBuilder.Label(k, v)
+	}
+
+	return arBuilder.Get()
+}
+
+func generateAuthorizationPolicySpec(rule gatewayv1beta1.Rule) *v1beta1.AuthorizationPolicy {
+	authorizationPolicySpec := builders.AuthorizationPolicySpecBuilder().
+		Selector(builders.SelectorBuilder().MatchLabels("app", *rule.Service.Name)).
```
videlov, line 38: we want to support matching service by `Namespace` as well, which is currently not quite straight forward with https://istio.io/latest/docs/reference/config/type/workload-selector/
another way ? or at least address it and tackle it later.

internal/processing/helpers.go
```diff
@@ -54,6 +55,18 @@ func setAccessRuleKey(hasPathDuplicates bool, rule rulev1alpha1.Rule) string {
 	return rule.Spec.Match.URL
 }
 
+func setAuthorizationPolicyKey(hasPathDuplicates bool, ap *istiosecurityv1beta1.AuthorizationPolicy) string {
```
videlov, line 58: shouldn't this method be called `getAuthorizationPolicyKey` , `set` is a bit confusing, same as the one above (which is old)?

internal/processing/helpers.go
```diff
@@ -54,6 +55,18 @@ func setAccessRuleKey(hasPathDuplicates bool, rule rulev1alpha1.Rule) string {
...
+	if hasPathDuplicates {
+		return fmt.Sprintf("%s:%s", ap.Spec.Rules[0].To[0].Operation.Paths, ap.Spec.Rules[0].To[0].Operation.Methods)
```
videlov, line 60: perhaps is a good idea to check for nils when digging into Spec, making it more resilient

internal/processing/processing.go
```diff
@@ -59,15 +60,26 @@ func (f *Factory) CalculateRequiredState(api *gatewayv1beta1.APIRule, config *he
...
 				ar = generateAccessRule(api, rule, rule.AccessStrategies, f.additionalLabels, f.defaultDomainName)
 			} else if config.JWTHandler == helpers.JWT_HANDLER_ISTIO {
 				//TODO generate based on config.JWTHandler="istio"
+				ap = generateAuthorizationPolicy(api, rule, f.additionalLabels)
+				ra = generateRequestAuthentication(api, rule, f.additionalLabels)
 			}
+
```
videlov: should test here for nil dereference on `ar`, `ap` and `ra`, before further process on them

internal/processing/processing.go
```diff
@@ -59,15 +60,26 @@ func (f *Factory) CalculateRequiredState(api *gatewayv1beta1.APIRule, config *he
 	var res State
 	pathDuplicates := hasPathDuplicates(api.Spec.Rules)
```
videlov, line 61: `pathDuplicates` suggests is a list of duplicates not a `bool`, could we refactor it to something more boolean like `hasPathDuplicates`?

internal/processing/processing.go
```diff
@@ -208,3 +246,81 @@ func (f *Factory) applyObjDiff(ctx context.Context, objToPatch *objToPatch) erro
...
+		if requiredState.accessRules[path] == nil {
+			objToDelete := &objToPatch{action: "delete", obj: rule}
+			arPatch[path] = objToDelete
+		}
+	}
+
+	return arPatch
+}
+
+func authorizationPolicyPatch(apPatch map[string]*objToPatch, actualState, requiredState *State) map[string]*objToPatch {
```
videlov: it applies all changes to the referenced `apPatch map[string]*objToPatch` returning it from the method is redundant, do we really need to return it?

internal/processing/processing.go
```diff
@@ -208,3 +246,81 @@ func (f *Factory) applyObjDiff(ctx context.Context, objToPatch *objToPatch) erro
...
+		if requiredState.authorizationPolicies[path] == nil {
+			objToDelete := &objToPatch{action: "delete", obj: authorizationPolicy}
+			apPatch[path] = objToDelete
+		}
+	}
+
+	return apPatch
+}
+
+func requestAuthenticationPatch(raPatch map[string]*objToPatch, actualState, requiredState *State) map[string]*objToPatch {
```
videlov: same here, as well as for the old one: ```func accessRulePatch(arPatch map[string]*objToPatch, actualState, requiredState *State) map[string]*objToPatch {```

videlov (changes requested):
manual and unit testing still pending

internal/processing/processing.go
videlov: yes, totally, agree I wanted to do it with switch as last commit from previous PR, but forgot it in the end, lets do it with `switch`

internal/processing/processing.go
```diff
@@ -169,7 +191,7 @@ func (f *Factory) CalculateDiff(requiredState *State, actualState *State) *Patch
 		vsPatch.obj = requiredState.virtualService
 	}
 
-	return &Patch{virtualService: vsPatch, accessRule: arPatch}
+	return &Patch{virtualService: vsPatch, accessRule: arPatch, authorizationPolicy: apPatch}
```
videlov, line 194:
```suggestion
	return &Patch{virtualService: vsPatch, accessRule: arPatch, authorizationPolicy: apPatch, requestAuthentication: raPatch}
```

videlov (changes requested):

internal/processing/authorization_policy_helpers.go
videlov: what about adding the AP's and RA's in same namespace as the service ? If I understand this correctly from the documentation: `The scope of label search is restricted to the configuration namespace in which the resource is present.`

videlov, line 38: `rule.Service` may be nil and we need to fallback to `Spec.service.Name`

videlov (changes requested):

internal/processing/authorization_policy_helpers.go
```diff
@@ -0,0 +1,45 @@
...
+
+	for k, v := range additionalLabels {
+		arBuilder.Label(k, v)
+	}
+
+	return arBuilder.Get()
+}
+
+func generateAuthorizationPolicySpec(rule gatewayv1beta1.Rule) *v1beta1.AuthorizationPolicy {
+	authorizationPolicySpec := builders.AuthorizationPolicySpecBuilder().
```
barchw, line 37: Since we want to support Ory rules access in parallel, we need to add something similar to this (probably as a separate AP resource):
```yaml
---
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: httpbin-allow-oathkeeper-sa
  namespace: default
spec:
  selector:
    matchLabels:
      app: httpbin
  action: ALLOW
  rules:
    - from:
        - source:
            principals:
              - "cluster.local/ns/kyma-system/sa/ory-oathkeeper"
    - to:
        - operation:
            hosts:
              - "httpbin*"
``` 
I'm not entirely sure however whether this implementation is secured from the point of SSRF, so I leave it as a suggestion, there might be some better way to do this

[Reference POC](https://github.tools.sap/kyma/backlog/issues/2528#issuecomment-1021635)

internal/processing/authorization_policy_helpers.go
```diff
@@ -0,0 +1,45 @@
...
+		Label(OwnerLabelv1alpha1, fmt.Sprintf("%s.%s", api.ObjectMeta.Name, api.ObjectMeta.Namespace))
+
+	for k, v := range additionalLabels {
+		arBuilder.Label(k, v)
+	}
+
+	return arBuilder.Get()
+}
+
+func generateAuthorizationPolicySpec(rule gatewayv1beta1.Rule) *v1beta1.AuthorizationPolicy {
```
barchw, line 36: Since we want to support Ory rules access in parallel, we need to add something similar to this (probably as a separate AP resource):
```yaml
---
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: httpbin
  namespace: foo
spec:
  action: ALLOW
  rules:
    - from:
        - source:
            principals:
              - "cluster.local/ns/kyma-system/sa/oathkeeper-maester-account"
```

videlov:
We also need Status update changes, see: https://github.com/kyma-project/api-gateway#additional-information
Have to apply status for APs and RAs: https://github.com/kyma-project/api-gateway/blob/d8d574c236d9f6422df710476508c195a4434aba/controllers/apirule_controller.go#L195-L202

internal/processing/authorization_policy_helpers.go
```diff
@@ -0,0 +1,52 @@
...
+	for k, v := range additionalLabels {
+		arBuilder.Label(k, v)
+	}
+
+	return arBuilder.Get()
+}
+
+func generateAuthorizationPolicySpec(api *gatewayv1beta1.APIRule, rule gatewayv1beta1.Rule) *v1beta1.AuthorizationPolicy {
+	var serviceName string
+	if rule.Service != nil && rule.Service.Name != nil {
```
triffer, line 38: I think we can omit the nil check for Service.Name as the field is not optional and the validation should fail.

internal/processing/authorization_policy_helpers.go
```diff
@@ -0,0 +1,52 @@
...
+		serviceName = *rule.Service.Name
+	} else {
+		serviceName = *api.Spec.Service.Name
```
triffer, line 41: We assume that the Spec.Service is set if no Service is set at Rule level, but as far as I know there is no validation that any of them are set. Do we need to add an additional check here?

internal/processing/processing.go
```diff
@@ -57,44 +58,50 @@ type CorsConfig struct {
...
+				ap = generateAuthorizationPolicy(api, rule, f.additionalLabels)
+				ra = generateRequestAuthentication(api, rule, f.additionalLabels)
+				res.authorizationPolicies[getAuthorizationPolicyKey(hasPathDuplicates, ap)] = ap
+				res.requestAuthentications[getRequestAuthenticationKey(ra)] = ra
 			}
-			res.accessRules[setAccessRuleKey(pathDuplicates, *ar)] = ar
 		}
 	}
 
 	//Only one vs
```
triffer, line 85: Perhaps we can remove this comment as it does not contain much information.

internal/processing/request_authentication_helpers.go
```diff
@@ -0,0 +1,50 @@
...
+		arBuilder.Label(k, v)
+	}
+
+	return arBuilder.Get()
+}
+
+func generateRequestAuthenticationSpec(api *gatewayv1beta1.APIRule, rule gatewayv1beta1.Rule) *v1beta1.RequestAuthentication {
+
+	var serviceName string
+	if rule.Service != nil && rule.Service.Name != nil {
```
triffer, line 39: I think we can omit the nil check for Service.Name as the field is not optional and the validation should fail.

triffer (dismissed):
I saw that we have a lot of checks for the JWT handler. I think it would make sense to have an abstraction for JWT Handlers and implement a component for each handler and pass that around instead of adding the if-checks. This way removing the support for a JWT handler would also be easier. What do you think?

videlov:
> I saw that we have a lot of checks for the JWT handler. I think it would make sense to have an abstraction for JWT Handlers and implement a component for each handler and pass that around instead of adding the if-checks. This way removing the support for a JWT handler would also be easier. What do you think?

Yes, this is a bit of unfortunate. Tbh we did it to avoid a lot of code duplication. @strekm did also raise this. I would vote to leave it like this for now and completely refactor it, once we implement a new handler. Or plan for it in for the future.

videlov:
Looks good for me, only fix the comments from Tim.

internal/processing/authorization_policy_helpers.go
barchw: Tracked in #75

triffer (approved):

videlov (approved):

PR #55, opened 2022-10-26
Make Cert Handler CRD Configurable

muralov:
* Make CRD configurable so that the app can generate for different CRDs
* Adap kustomize resources to use CRD name

**Description**
Make CRD configurable so that cert-handler can be configured for a different CRD too

- Make CRD configurable so that the app can generate for different CRDs
- Adap kustomize resources to use CRD name

**Related issue(s)**
#15346


dariusztutaj (approved):
/meow

PR #57, opened 2022-10-27
Provide make target to generate CRD with patches

cnvergence:
Make it easier to maintain crds in different projects
- https://github.com/kyma-project/kyma/blob/main/installation/resources/crds/api-gateway/apirules.gateway.crd.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,6 +173,11 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
+# Generate CRD with patches
+.PHONY: generate-crd-yaml
+generate-crd-yaml: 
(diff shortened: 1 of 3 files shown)


barchw:
I would say this is unnecessary, as the conversion is applied as kustomization patch in https://github.com/kyma-project/api-gateway/blob/main/config/crd/patches/webhook_in_apirules.yaml

cnvergence:
> I would say this is unnecessary, as the conversion is applied as kustomization patch in https://github.com/kyma-project/api-gateway/blob/main/config/crd/patches/webhook_in_apirules.yaml

Alright, shall we then generate and keep the outcome CRD here?

barchw (approved):

Issue #40, opened 2022-09-23
Feature toggle for Istio based jwt handler


strekm:
**Description**

Implement feature flag (named e.g. istioJwt) for handling istio based jwt handler. Feature flag should be passed to api-gateway controller as environment variable. Default values should be false. Feature toggle should control behaviour of jwt handler (changing logic of jwt handler is out of scope). Default false value of feature toggle means that jwt handler implementation is based on ORY oathkeeper, true value switches behaviour to jwt implementation based on Istio (implementation out of scope). 

**Reasons**

Support for early implementation of jwt handler based on Istio

**Attachments**

Part of: https://github.com/kyma-project/api-gateway/issues/30

PR #77, opened 2022-11-04
Config fail early

videlov:
**Description**

- default to `ory` if ConfigMap is missing
- move config validation in controller `reconcile()` before actually checking if is a new APIRule
- add watcher for the config map to APIRule controller (update event case)
- do not crash api-gateway pod, just log an error in controller logs and also set status of reconciled APIRule to error

**Related issue(s)**

https://github.com/kyma-project/api-gateway/issues/40

### DIFF
--- a/controllers/apirule_controller.go
+++ b/controllers/apirule_controller.go
@@ -101,6 +101,7 @@ func (p configMapPredicate) Generic(e event.GenericEvent) bool {
 //+kubebuilder:rbac:groups=security.istio.io,resources=authorizationpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=security.istio.io,resources=requestauthentications,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 
 func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	r.Log.Info("Starting reconcilation")
--- a/controllers/apirule_controller.go
+++ b/controllers/apirule_controller.go
@@ -63,13 +63,30 @@ type APIRuleReconciler struct {
 //+kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list;watch;create;update;patch;delete
 
 func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	//_ = r.Log.WithValues("Api", req.NamespacedName)
-
 	r.Log.Info("Starting reconcilation")
 
 	api := &gatewayv1beta1.APIRule{}
+	config, err := helpers.LoadConfig()
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf(`Config loading failure {"controller": "Api", "request": "%s/%s", "file": %s}, will use default config`, api.Namespace, api.Name, helpers.CONFIG_FILE))
+		config = helpers.DefaultConfig()
+	}
 
-	err := r.Client.Get(ctx, req.NamespacedName, api)
+	validator := validation.APIRule{
+		ServiceBlockList:  r.ServiceBlockList,
+		DomainAllowList:   r.DomainAllowList,
+		HostBlockList:     r.HostBlockList,
+		DefaultDomainName: r.DefaultDomainName,
+	}
+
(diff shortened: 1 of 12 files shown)


controllers/apirule_controller.go
```diff
@@ -148,6 +182,8 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 func (r *APIRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gatewayv1beta1.APIRule{}).
+		For(&corev1.ConfigMap{}).
```
barchw, line 185:
```suggestion
		Owns(&corev1.ConfigMap{}).
```

barchw (changes requested):

controllers/apirule_controller.go
```diff
@@ -80,31 +109,36 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	}
 
-		//1.1) Get the configuration
-		config, err := helpers.LoadConfig()
-		if err != nil {
-			//If configuration is not available not been able to continue.
-			r.Log.Info(fmt.Sprintf(`Configuration loading failed {"controller": "Api", "request": "%s/%s", "file": %s}`, api.Namespace, api.Name, helpers.CONFIG_FILE))
-			return r.setStatusForError(ctx, api, err, gatewayv1beta1.StatusError)
-		}
+	config, err := helpers.LoadConfig()
```
barchw, line 119: We can check that the received req contains a ConfigMap, and then proceed to update if needed, instead of running the LoadConfig every loop

barchw (changes requested):

controllers/apirule_controller.go
```diff
@@ -63,12 +95,9 @@ type APIRuleReconciler struct {
 //+kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list;watch;create;update;patch;delete
 
 func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	//_ = r.Log.WithValues("Api", req.NamespacedName)
-
 	r.Log.Info("Starting reconcilation")
 
 	api := &gatewayv1beta1.APIRule{}
-
 	err := r.Client.Get(ctx, req.NamespacedName, api)
```
barchw, line 101: This will probably produce an error on ConfigMap event

internal/helpers/config.go
```diff
@@ -17,6 +17,10 @@ type Config struct {
 	JWTHandler string `yaml:"jwtHandler"`
 }
 
+func DefaultConfig() *Config {
+	return &Config{JWTHandler: JWT_HANDLER_ORY}
+}
+
 func LoadConfig() (*Config, error) {
 	configData, err := ReadFileHandle(CONFIG_FILE)
```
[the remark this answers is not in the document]
videlov: I would still keep the mount, since it is more reliable and always there in container. Eventing could lead to some race-condition issues.

controllers/apirule_controller.go
videlov: Yes, we could optimize that, although small one.

controllers/apirule_controller.go
videlov: yes, this will produce error. will fix it, need to check what resource/request is being reconciled.

internal/helpers/config.go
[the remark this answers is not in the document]
barchw: I meant, that the ConfigMap can be passed as a parameter from the reconcile loop. The `update` `create` `delete` event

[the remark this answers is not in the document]
videlov: But what happens if we get a reconcile run for APIRule first before ConfigMap being passed ?

controllers/apirule_controller.go
```diff
@@ -148,6 +201,8 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 func (r *APIRuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gatewayv1beta1.APIRule{}).
+		Owns(&corev1.ConfigMap{}).
+		WithEventFilter(&configMapPredicate{Log: r.Log}).
```
barchw, line 205: This seems to be preventing the controller from reconciling on apirule creation, as this is applied to all watched resources
```suggestion
```

barchw (changes requested):

controllers/apirule_controller.go
videlov: Fixed that, but still don't get reconciles for CM updates, looking into it.

barchw (approved):
/meow

Issue #95, opened 2022-11-15
Refactor processing of handlers


strekm:
**Description**

Introducing istio based jwt handler in parallel to oathkeeper handler exposed need to refactor source code in `processing.go` file. Break processing of handlers into separate files including unit tests. 

ACs:
- [x] slimmer `processing.go`
- [x] introduced structure should support adding handlers
- [x] It is possible to have different object kinds processed based on the handler
- [x] It is possible to have different processing logic of an object based on the handler
- [x] unit tests are not too sensitive to changes
- [x] component tests and manual tests are green

**Reasons**

Simplifying source code and improving maintenance

**DoD:**
~~- [ ] provide documentation~~
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~
- [x] provide unit tests
- [x] test on SKR
- [x] verify [Get started guides](https://kyma-project.io/docs/kyma/latest/02-get-started/) 
~~- [ ] verify resource limits~~
- [ ] followup issue
~~- [ ] create release and bump in kyma~~

triffer:
**PR:**
- https://github.com/kyma-project/api-gateway/pull/99

PR #99, opened 2022-11-17
Processing restructure

triffer:
**Description**

- Restructure processing to make it to support easier dynamic configuration for different handlers

**Related issue(s)**
- https://github.com/kyma-project/api-gateway/issues/95


barchw (changes requested):

internal/processing/ory/jwt_validator.go
```diff
@@ -0,0 +1,57 @@
+package ory
```
barchw, line 1: I would move this to validation package

internal/processing/istio/jwt_validator.go
```diff
@@ -0,0 +1,14 @@
+package istio
```
barchw, line 1: I would move this to validation package

internal/processing/ory/jwt_validator.go
triffer: If we move it to validation then the visibility would need to change and it could be used outside of the intended scope. 
That's why I'd rather think about a different structure for all packages as imho validation is part of the processing and might be better a sub-package and with a similar sub-package for each specific reconciliation e.g. ory/validation, istio/validation, processing/validation.

I think we also need to improve the implementation of validation at some point.

What do you think @barchw ?

barchw: Yes that would be ok, or validation/ory

barchw (changes requested):

internal/processing/types.go
```diff
@@ -0,0 +1,50 @@
+package processing
+
+import (
+	v1beta12 "istio.io/api/networking/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ObjectChange struct {
+	Action string
```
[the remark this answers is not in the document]
triffer: I thought about this and didn't know what would be the best approach in Go. In TS I'd have used a string literal type/union, but I'm not aware of something similar in Go.
What do you think would be the best option here?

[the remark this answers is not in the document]
barchw: I am aware only of the iota enum option and I don't think there is a way of directly mapping string to enum
```go
type Action uint64

const (
	Undefined Action = iota
	Create
	Update
	Delete
)

func (s Action) String() string {
	switch s {
	case Create:
		return "create"
	case Update:
		return "update"
	case Delete:
		return "delete"
	}
	return "unknown"
}
```

barchw:
This looks good in my opinion, maybe just adding some comments about what the interfaces and their function do would help a little bit with readability, but otherwise 👍

barchw (approved):
/woof

Issue #149, opened 2022-12-28
component test for istio based jwt apirule creation


strekm:
**Description**

Implement component test exposing httpbin with istio jwt apirule. Assuming cluster is created, kyma installed (only Istio is required) api-gateway controller is installed with FT on test should deploy httpbin workload and add it to service mesh. APIRule should be created with all possible fields (consider required scopes) and then httpbin endpoint called. 

ACs:
- [x] test deploys httpbin workload as a part of service mesh
- [x] test creates apirule of type istio jwt with all possible fields
- [x] test calles exposed endpoint
- [x] test should be gherkinized 

**Reasons**

automatisation of testing effort

**DoD:**
~~- [ ] provide documentation~~
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~
~~- [ ] provide unit tests~~
- [x] test on SKR
~~- [ ] verify [Get started guides](https://kyma-project.io/docs/kyma/latest/02-get-started/)~~
~~- [ ] verify resource limits~~
- [x] followup issue
~~- [ ] create release and bump in kyma~~

**PR**
- [x] https://github.com/kyma-project/api-gateway/pull/164

**Attachments**

PR #164, opened 2023-01-16
API Gateway integration Istio jwt test

videlov:
**Description**

- API Gateway component/integration Istio JWT test

**Related issue(s)**

https://github.com/kyma-project/api-gateway/issues/149


go.work
```diff
@@ -0,0 +1,6 @@
+go 1.19
```
barchw, line 1: Please add to .gitignore

barchw (changes requested):

tests/integration/istio_jwt_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func InitializeScenarioIstioJWT(ctx *godog.ScenarioContext) {
+	mainScenario, err := CreateScenario(istioJwtApiruleFile, "istio-jwt")
+	if err != nil {
+		t.Fatalf("could not initialize unsecure endpoint scenario err=%s", err)
+	}
+
+	scenario := istioJwtScenario{mainScenario}
+
+	ctx.Step(`^IstioJWT: There is an deployment secured with JWT on path /ip$`, scenario.thereIsAnEndpoint)
```
barchw, line 25:
```suggestion
	ctx.Step(`^IstioJWT: There is an deployment secured with JWT on path "([^"]*)"$`, scenario.thereIsAnEndpoint)
```

barchw (changes requested):

tests/integration/go.mod
```diff
@@ -0,0 +1,66 @@
+module github.com/kyma-incubator/api-gateway/tests/integration
```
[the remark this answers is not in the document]
videlov: I wanted to have it in a submodule and not mix it up with the api-gateway controller deliverable. Are you sure you want to have everything in one module ?

tests/integration/manifests/hydra-client.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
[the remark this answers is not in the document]
videlov: @strekm can you elaborate on this? since I was not specified and I do not remember what we have decided. Do we want to completely get away from ory.Hydra and use different JWT provider?

tests/integration/go.mod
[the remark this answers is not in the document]
barchw: I think that's the recommendation for Go modules so it might be the expected configuration for some tools, but if you feel that would be better since we only use it for tests it can stay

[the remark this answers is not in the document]
videlov: I think I like it to be separate. Will see how it looks in a single module and adapt it if it looks better.

tests/integration/pkg/jwt/config.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewJwtConfig() (Config, error) {
+	env := envConfig{}
+	err := envconfig.Init(&env)
+	if err != nil {
+		return Config{}, errors.Wrap(err, "while loading environment variables")
+	}
+
+	config := Config{EnvConfig: env}
+	config.ClientConfig = clientConfig{ClientTimeout: env.ClientTimeout}
```
barchw, line 32:
```suggestion
	config.ClientConfig = clientConfig(env)
```

tests/integration/manifests/global-commons.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "{{.Namespace}}"
+  labels:
+    istio-injection: enabled
+---
+apiVersion: v1
```
barchw, line 8: I think this Secret is not needed with Istio JWT

tests/integration/test_helpers.go
```diff
@@ -0,0 +1,482 @@
...
+	goDogOpts.Paths = pflag.Args()
+	if os.Getenv(exportResultVar) == "true" {
+		goDogOpts.Format = "pretty,junit:junit-report.xml,cucumber:cucumber-report.json"
+	}
+
+	if err := envconfig.Init(&conf); err != nil {
+		log.Fatalf("Unable to setup config: %v", err)
+	}
+
+	if conf.IsMinikubeEnv {
```
barchw, line 129: We can probably remove this as we are using K3D

tests/integration/test_helpers.go
```diff
@@ -0,0 +1,482 @@
...
+func SetupCommonResources(namePrefix string) {
+	namespace = fmt.Sprintf("%s-%s", namePrefix, generateRandomString(6))
+	randomSuffix6 := generateRandomString(6)
+	oauthSecretName := fmt.Sprintf("%s-secret-%s", namePrefix, randomSuffix6)
+	oauthClientName := fmt.Sprintf("%s-client-%s", namePrefix, randomSuffix6)
+	log.Printf("Using namespace: %s\n", namespace)
+	log.Printf("Using OAuth2Client with name: %s, secretName: %s\n", oauthClientName, oauthSecretName)
+	oauth2Cfg = &clientcredentials.Config{
+		ClientID:     conf.ClientID,
+		ClientSecret: conf.ClientSecret,
```
barchw, line 179: This OAuth2Client and Secret generation is not needed

barchw (changes requested):

tests/integration/istio_jwt_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func InitializeScenarioIstioJWT(ctx *godog.ScenarioContext) {
+	mainScenario, err := CreateScenario(istioJwtApiruleFile, "istio-jwt")
+	if err != nil {
+		t.Fatalf("could not initialize unsecure endpoint scenario err=%s", err)
+	}
+
+	scenario := istioJwtScenario{mainScenario}
+
+	ctx.Step(`^IstioJWT: There is an deployment secured with JWT on path "([^"]*)"$`, scenario.thereIsAnEndpoint)
```
barchw, line 25:
```suggestion
	ctx.Step(`^IstioJWT: There is a deployment secured with JWT on path "([^"]*)"$`, scenario.thereIsAnEndpoint)
```

tests/integration/test_helpers.go
```diff
@@ -169,15 +154,11 @@ func InitTestSuite() {
...
-	oauthSecretName := fmt.Sprintf("%s-secret-%s", namePrefix, randomSuffix6)
-	oauthClientName := fmt.Sprintf("%s-client-%s", namePrefix, randomSuffix6)
 	log.Printf("Using namespace: %s\n", namespace)
-	log.Printf("Using OAuth2Client with name: %s, secretName: %s\n", oauthClientName, oauthSecretName)
 	oauth2Cfg = &clientcredentials.Config{
 		ClientID:     conf.ClientID,
 		ClientSecret: conf.ClientSecret,
-		TokenURL:     fmt.Sprintf("%s/oauth2/token", conf.IasAddr),
+		TokenURL:     fmt.Sprintf("%s/oauth2/token", conf.IssuerUrl),
 		Scopes:       []string{"read"},
```
barchw, line 162: Seems like most of those parameters aren't used right now

tests/integration/test_helpers.go
```diff
@@ -191,12 +172,10 @@ func SetupCommonResources(namePrefix string) {
 		Namespace         string
 		OauthClientSecret string
```
barchw, line 173: Those aren't used as well

barchw (changes requested):

tests/integration/test_helpers.go
videlov: These are for obtaining a token within the test scenario (see https://github.com/kyma-project/api-gateway/blob/60ef392d94ac6b7f451fedd546138fd5775457c7/tests/integration/pkg/jwt/auth.go#L29). I did remove all other unused. Hopefully got them all.

barchw: Is `Scopes` really needed?

videlov: Ah, you meant scopes, perhaps will work without it. Will test.

barchw (approved):

Issue #84, opened 2022-11-09
Proper handling of APIRule Status


strekm:
**Description**

Status object of APIRule should be adapted to changes introduced by [istio based jwt handler](https://github.com/kyma-project/api-gateway/issues/34). Existing logic should take into consideration sub resources created by different handlers. Status of subresource should be added only if that subresource will be processed. 

ACs:
- [x] status of subresource is only showed if that subresource will be managed by handler
- [x] status of istio subresources (RA & AP) is only shown if FT is turned on 

**Reasons**

Introducing jwt based handler

**DoD:**
- [x] provide unit tests
- ~~followup issue~~
- [x] execute manual tests

**Attachments**
part of: https://github.com/kyma-project/api-gateway/issues/30

**PR**
- https://github.com/kyma-project/api-gateway/pull/158

barchw:
We need to update APIRule CRD with AuthorizationPolicy and RequestAuthentication status

PR #166, opened 2023-01-17
Add new file for status integration tests and restructure variables

triffer:
**Description**

- Add integration test for changes to status

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/84


barchw (changes requested):
Small remarks, either than that LGTM
I would only suggest to add a test with status "Error" on one of the sub resources

barchw (approved):

PR #183, opened 2023-01-25
Add Istio JWT doc

werdes72:


docs/istio-jwt.md
```diff
@@ -0,0 +1,52 @@
+# API-Gateway JWT Handler Feature Toggle
+
+## Overview
+
+We support two JWT handlers in API-Gateway at the moment. Switching between them is configurable via the following ConfigMap: `kyma-system/api-gateway-config`
+
+## Swithing between JWT handlers
```
triffer, line 7:
```suggestion
## Switching between JWT handlers
```

docs/istio-jwt.md
```diff
@@ -0,0 +1,52 @@
...
+          config:
+            authentications:
+            - issuer: $ISSUER
+              jwksUri: $JWKS_URI
+            authorizations:
+            - requiredScopes: ["test"]
+            - requiredScopes: ["read", "write"]
+```
+
+The `authorizations` field defined above will require your JWT to contain either `test` scope OR `read` AND `write` scopes. The scope value has to be in one of the following keys: `scp`, `scope`, `scopes`, to ensure backwards compatibility with ory oathkeeper.
```
triffer, line 52: I think we should add a table with the fields specific for the Istio JWT handler like [here](https://github.com/kyma-project/api-gateway#sample-custom-resource).
What do you think?

docs/istio-jwt.md
```diff
@@ -0,0 +1,52 @@
+# API-Gateway JWT Handler Feature Toggle
```
triffer, line 1: Since we have docs for Istio and Ory JWT handler inside it might be better to rename the file to `jwt-handler.md`?

werdes72: Hmm, the only thing with ory here is how to use feature toggle to enable it, but the APIRule usage is related to istio only.

triffer (approved):

PR #192, opened 2023-01-31
Add Istio JWT tests with scopes

werdes72:


tests/integration/test_helpers.go
```diff
@@ -49,7 +49,8 @@ const (
 	manifestsDirectory        = "manifests/"
 	testingAppFile            = "testing-app.yaml"
 	globalCommonResourcesFile = "global-commons.yaml"
-	istioJwtApiruleFile       = "istio-jwt-strategy.yaml"
+	istioJwtApiRuleFile       = "istio-jwt-strategy.yaml"
+	istioJwtApiRuleScopesFile = "istio-jwt-scopes.yaml"
```
triffer, line 53: This file is only used in `istio_jwt_scopes_test.go`so it might be better to have it inside of the helper as there is no benefit in doing this.

tests/integration/istio_jwt_scopes_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (o *istioJwtScopesScenario) callingTheEndpointWithInvalidTokenShouldResultInStatusBetween(path string, lower, higher int) error {
+	return helper.CallEndpointWithHeadersWithRetries(anyToken, authorizationHeaderName, fmt.Sprintf("%s%s", o.url, path), &helpers.StatusPredicate{LowerStatusBound: lower, UpperStatusBound: higher})
+}
+
+func (o *istioJwtScopesScenario) callingTheEndpointWithoutTokenShouldResultInStatusBetween(path string, lower, higher int) error {
+	return helper.CallEndpointWithRetries(fmt.Sprintf("%s%s", o.url, path), &helpers.StatusPredicate{LowerStatusBound: lower, UpperStatusBound: higher})
+}
+
+func (o *istioJwtScopesScenario) callingTheSecondEndpointWithValidTokenWithoutScopesShouldResultInStatusBetween(path, tokenType string, lower, higher int) error {
```
triffer, line 58: From what I understand only this function is adding new testing capabilities for scopes as the other functions already exist for istio JWT.

tests/integration/main_test.go
```diff
@@ -46,6 +50,35 @@ func TestIstioJwt(t *testing.T) {
 	}
 }
 
+func TestIstioJwtScopes(t *testing.T) {
```
triffer, line 53: We have a lot of duplicated code and it might improve readbility if we have a function that will take care of setup, cleanUp and run a given JWT test suite as there is some common setup.

tests/integration/features/istio-jwt/istio_jwt_scopes.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Exposing two endpoints with Istio JWT authorization strategy with scopes
+
+  Scenario: IstioJWTScopes: Calling an endpoint secured with JWT with a valid token
+    Given IstioJWTScopes: There is a deployment secured with JWT on path "/ip"
+    Then IstioJWTScopes: Calling the "/ip" endpoint without a token should result in status between 400 and 403
```
triffer, line 5: The unhappy steps that validate general JWT behaviour are a already tested in IstioJWT feature. I think we should not duplicate them.

tests/integration/features/istio-jwt/istio_jwt_scopes.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Exposing two endpoints with Istio JWT authorization strategy with scopes
+
+  Scenario: IstioJWTScopes: Calling an endpoint secured with JWT with a valid token
```
triffer, line 3: Is it possible to have the scenario description more specific? Something like "Calling endpoints secured with JWT and enforcing JWT scope claims"?

tests/integration/features/istio-jwt/istio_jwt_scopes.feature
```diff
@@ -0,0 +1,8 @@
...
+    Given IstioJWTScopes: There is a deployment secured with JWT on path "/ip"
```
triffer, line 4: More information for the given endpoints might it easier to understand the test without reading implementation, e.g. `There is one deployment that requires scope claims "read", ... and there is a deployment that requires scope claims "test", ...`
We can then also adapt the assertion text adding more info, e.g. `Calling the "/ip" endpoint with a valid "JWT" token and scope claims "read", "test" and "write"  should result in status between 200 and 299`.
Maybe it makes sense to have two scenarios one for happy and one for unhappy tests.

triffer (commented):
We need some specific setup of JWT scopes to work. I think we need to document this, because we can't have this automated in this repository.
In my case a `cucumber-report.json` and `junit-report.xml` were also created locally, which we may want to add to the gitignore file.

videlov (approved):

Issue #181, opened 2023-01-24
Improve handling of Authorization Policy Spec changes


triffer:
**Description**

During the review of [Support for required scopes in Istio based jwt handler](https://github.com/kyma-project/api-gateway/issues/148) we found out that the implemented logic for deciding whether a Authorization Policy needs to be created, updated or deleted can't be easily extended.
The reason for this is that we added the support for scopes and we now create a new AP for each Authorization configured in the handler config. With the current way of creating unique keys, that means we would need to create a unique key for each Authorization that includes all rules and conditions.
Since some of these values are arrays, we always have to arrange them in the same way so that we can compare them.
The proposed solution was to use a hashing of the Authorization Policy Spec to make it comparable.
As this is a bigger change, we want to handle this as a separate issue. In the meantime, to avoid resources not being created or being in the wrong state, we decided to always create new authorization policies and delete the existing ones.

Also the current implementation to detect Authorization Policy resource changes will run an update for unchanged resources as we don't really verify if it has changed or not. We only know that it was not deleted and doesn't need to be created, so run an update to make sure it is in the right place.

**Reasons**
Since the re-creation and unnecessary updates of policies might affect the performance or cause issues with dependencies like Envoy we should evaluate a better solution. A possible solution could include using labels and creating a hash from the Authorization Policy Spec.

**AC**
- [x] The integration test "should create and update authorization policies when adding new authorization" passes
- [x] Authorization policies are only deleted when they are no longer required by the ApiRule configuration
~~- [ ] Authorization policies are only updated when they contain changes~~ 
- [x] The solution should be generic so it can also be used for other resources, e.g. RequestAuthentication

**DoD:**
- [x] provide documentation
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~
- [x] provide unit tests
- [x] test on ~~SKR~~ Gardener
~~- [ ] verify [Get started guides](https://kyma-project.io/docs/kyma/latest/02-get-started/)~~
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] create release and bump in kyma~~

**Attachments**
- part of https://github.com/kyma-project/api-gateway/issues/30
- related to https://github.com/kyma-project/api-gateway/issues/148

- https://github.com/kyma-project/api-gateway/pull/225

triffer:
Removed AC "Authorization policies are only updated when they contain changes" as we should always call the update to make sure that the AP is in the expected state if there were manual changes.

PR #225, opened 2023-02-24
Add hash to Authorization Policy

barchw:
**Description**

- Hash AP spec
- Use the hash to identify the AP
- Add indexing to APs
- Use yaml array order for deterministic update of AP corresponding to Authorizations array entry

**Related issue(s)**

- #181


internal/processing/processors/authorization_policy_processor.go
```diff
@@ -34,34 +40,101 @@ func (r AuthorizationPolicyProcessor) EvaluateReconciliation(ctx context.Context
...
-func (r AuthorizationPolicyProcessor) getObjectChanges(desiredAps []*securityv1beta1.AuthorizationPolicy, actualAps []*securityv1beta1.AuthorizationPolicy) []*processing.ObjectChange {
+func (r AuthorizationPolicyProcessor) getObjectChanges(desiredAps map[string][]*securityv1beta1.AuthorizationPolicy, actualAps map[string]map[string]*securityv1beta1.AuthorizationPolicy) []*processing.ObjectChange {
 	var apObjectActionsToApply []*processing.ObjectChange
 
-	for _, ap := range desiredAps {
-		objectAction := processing.NewObjectCreateAction(ap)
-		apObjectActionsToApply = append(apObjectActionsToApply, objectAction)
+	for hashTo, toDesiredAPs := range desiredAps {
+		for _, ap := range toDesiredAPs {
+			// As both the order of Authorizations and APs is static we can update them according to array index
```
barchw, line 103: This resolves the problem of having APIRule configuration like this one:
```yaml
- accessStrategies:
      - config:
          authentications:
          - issuer: https://kymagoattest.accounts400.ondemand.com
            jwksUri: https://kymagoattest.accounts400.ondemand.com/oauth2/certs
          authorizations:
          - audiences:
            - https://example.com
            requiredScopes:
            - read
          - requiredScopes:
            - write
        handler: jwt
```
The problem with this configuration is that if both of the scopes would change, we cannot be sure with the hashcode that update is executed on the AP corresponding to e.g. `read` scope, which may cause service downtime, as we would have to APs that are exposing the same path at the same time, and the AP for `write` would not be present

internal/processing/processors/authorization_policy_processor.go
```diff
@@ -34,34 +40,101 @@ func (r AuthorizationPolicyProcessor) EvaluateReconciliation(ctx context.Context
...
 	var apList securityv1beta1.AuthorizationPolicyList
 	if err := client.List(ctx, &apList, ctrlclient.MatchingLabels(labels)); err != nil {
 		return nil, err
 	}
 
-	authorizationPolicies := append([]*securityv1beta1.AuthorizationPolicy{}, apList.Items...)
+	authorizationPolicies := make(map[string]map[string]*securityv1beta1.AuthorizationPolicy)
+	for i, ap := range apList.Items {
+		index, ok := ap.Labels[processing.IndexLabelName]
+		if !ok {
```
triffer, line 62: Isn't this more or less a migration case? I think we can ignore something like this and just re-create APs in this case for now as it should happen only once for currently existing APs.

barchw: If we don't store the APs without labels in some way, they won't be deleted and this will result in AP duplication. Technically this should only be the case right now so I can delete this but I'm not 100% sure if it is OK to have them duplicated.

barchw: That being said, I think that the `else if` statement at line 110 should be OK to delete, as this will only cause the APs to be recreated

barchw: https://github.com/kyma-project/api-gateway/pull/225/commits/5fd017c1b88a4402982f9ee960922ce71176844f

triffer (approved):

Issue #274, opened 2023-03-24
Changing JWT handler through config map not working reliable


triffer:
AC:
- [x] Test behaviour on Azure

**Description**

When changing the JWT handler through the config map `api-gateway-config` the reconciler won't pick up the newly changed handler. Only a restart of the controller pod will make the reconciler use the updated handler.
This means a change in the config map no longer triggers the reconciliation.

It seems that this is related to the change in https://github.com/kyma-project/api-gateway/pull/193.

**PR**
- https://github.com/kyma-project/api-gateway/pull/275

**Expected result**

Updating the `api-gateway-config` config map will change the configured handler without the need of restarting the pod.

**Actual result**

After updating the `api-gateway-config` config map the old handler is still used for reconciliation.

**Steps to reproduce**

- Set api-gateway-config CM to ory and deploy api-gateway-controller 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: api-gateway-config
  namespace: kyma-system
data:
  api-gateway-config: 'jwtHandler: ory'
```

- Create APIRule with Ory-based jwt config, this will create APIRule with Status OK
```
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: my-rule
  namespace: kyma-system
spec:
  gateway: kyma-system/kyma-gateway
  host: host.local.kyma.dev
  service:
    name: service
    port: 8000
  rules:
    - path: /headers
      methods: [ "GET" ]
      accessStrategies:
        - handler: jwt
          config:
            jwks_urls:
              - https://some.url.com/oauth2/certs
```

- Change api-gateway-config CM to istio 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: api-gateway-config
  namespace: kyma-system
data:
  api-gateway-config: 'jwtHandler: istio'
```

- Update APIRule to Istio-based jwt config will result in APIRule in Error state with error `Validation error: Attribute ".spec.rules[0].accessStrategies[0].config.authentications": Configuration for authentications is not supported with Ory handler`
```
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: my-rule
  namespace: kyma-system
spec:
  gateway: kyma-system/kyma-gateway
  host: host.local.kyma.dev
  service:
    name: service
    port: 8000
  rules:
    - path: /headers
      methods: [ "GET" ]
      accessStrategies:
        - handler: jwt
          config:
            authentications:
[...]

PR #275, opened 2023-03-28
Fix changing JWT handler with config map

triffer:
**Description**
The mechanism to change the JWT handler used by api-gateway is not working during runtime. To fix this we need to adapt the EventFilter config for the controller.

- Adapt the event filter config that the config map change will be reconciled by the controller
- Improve integration tests, so they are using the "real" controller config, so we are able to detect such problems earlier

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/274


controllers/apirule_controller.go
```diff
@@ -71,35 +72,34 @@ const (
...
 	predicate.Funcs
 }
 
-func (p configMapPredicate) Create(e event.CreateEvent) bool {
+func (p isApiGatewayConfigMapPredicate) Create(e event.CreateEvent) bool {
 	return p.Generic(event.GenericEvent(e))
 }
 
-func (p configMapPredicate) DeleteFunc(e event.DeleteEvent) bool {
+func (p isApiGatewayConfigMapPredicate) DeleteFunc(e event.DeleteEvent) bool {
```
barchw, line 84:
```suggestion
func (p isApiGatewayConfigMapPredicate) Delete(e event.DeleteEvent) bool {
```

controllers/apirule_controller.go
```diff
@@ -232,37 +232,40 @@ func (r *APIRuleReconciler) getReconciliation(config processing.ReconciliationCo
...
+		// We need to filter for generation changes, because we had an issue that on Azure clusters the APIRules were constantly reconciled.
+		For(&gatewayv1beta1.APIRule{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(&isApiGatewayConfigMapPredicate{Log: r.Log})).
 		Complete(r)
 }
 
 // Updates api status. If there was an error during update, returns the error so that entire reconcile loop is retried. If there is no error, returns a "reconcile success" value.
 func (r *APIRuleReconciler) updateStatusOrRetry(ctx context.Context, api *gatewayv1beta1.APIRule, status processing.ReconciliationStatus) (ctrl.Result, error) {
-	_, updateStatusErr := r.updateStatus(ctx, api, status)
+	_, updateStatusErr := r.updateStatus(ctx, api, status, &r.Log)
```
barchw, line 243: The logger is a field in `APIRuleReconciler`. Why do we pass it in function call?

controllers/apirule_controller.go
```diff
@@ -278,7 +281,7 @@ func retryReconcile(err error) (ctrl.Result, error) {
 	return reconcile.Result{Requeue: true}, err
 }
 
-func (r *APIRuleReconciler) updateStatus(ctx context.Context, api *gatewayv1beta1.APIRule, status processing.ReconciliationStatus) (*gatewayv1beta1.APIRule, error) {
+func (r *APIRuleReconciler) updateStatus(ctx context.Context, api *gatewayv1beta1.APIRule, status processing.ReconciliationStatus, logger *logr.Logger) (*gatewayv1beta1.APIRule, error) {
```
barchw, line 284: The logger is a field in APIRuleReconciler. Why do we pass it in function call?

controllers/api_controller_integration_test.go
```diff
@@ -914,166 +765,151 @@ var _ = Describe("APIRule Controller", func() {
...
+		expectApiRuleStatus(apiRuleName, gatewayv1beta1.StatusError)
+
+		By("Verify VirtualService for APIRule is not created")
+		Eventually(func(g Gomega) {
+			vsList := networkingv1beta1.VirtualServiceList{}
+			g.Expect(c.List(context.TODO(), &vsList, apiRuleLabelMatcher)).Should(Succeed())
+			g.Expect(vsList.Items).To(HaveLen(0))
+		}, eventuallyTimeout).Should(Succeed())
+
+		By("Deleting existing VirtualService with duplicated host configuration")
```
barchw, line 899: This commenting grammar convention is inconsistent, we should either have every `By` in infinitive `-ing` form or in imperative `Verify`, `Create`

controllers/suite_test.go
```diff
@@ -165,19 +154,15 @@ var _ = BeforeSuite(func(specCtx SpecContext) {
 		Config:                 &helpers.Config{},
 
 		// Run the suite with period that won't interfere with tests
```
barchw, line 156: This comment is no longer valid
```suggestion
```

barchw (changes requested):

barchw (approved):
/woof

Issue #236, opened 2023-03-03
[Istio JWT] Allow exposing services with path /.*


barchw:
**Description**
We should allow exposing services with wildcard path `/.*`. Since VS supports RegexMatch on path and AP doesn't the configuration needs to take that into consideration.

**Configuration**
Istio VS supports regex match, so the path can be set as `/.*`. Since Istio AP only supports prefix, suffix and wildcard match we need to configure it with `*`

**DoD:**
~~- [ ] provide documentation~~
~~- [ ] publish release notes and _What's New_ updates for Kyma customers~~
- [x] provide unit tests
- [x] test on gardener

~~- [ ] verify [Get started guides](https://kyma-project.io/docs/kyma/latest/02-get-started/)~~
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~

**PR**
- [x] https://github.com/kyma-project/api-gateway/pull/300

PR #300, opened 2023-04-07
Allow exposure of endpoints with wildcard  `/*` and `/.*`  in Istio JWT

barchw:
**Description**

- If the path is `/.*` or `/*` set VS and AP path to `/*` with prefix match.

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/236


internal/processing/istio/authorization_policy_processor.go
```diff
@@ -149,6 +149,13 @@ func generateAuthorizationPolicySpec(api *gatewayv1beta1.APIRule, rule gatewayv1
 }
 
 func withTo(b *builders.RuleBuilder, rule gatewayv1beta1.Rule) *builders.RuleBuilder {
+	if rule.Path == "/.*" {
```
triffer, line 152:
```suggestion
        // APIRule and VirtualService supported a regex match. Since AuthorizationPolicy supports only prefix, suffix and wildcard 
        // and we have clusters with "/.*" in APIRule, we need special handling of this case.
	if rule.Path == "/.*" {
```

PR #303, opened 2023-04-11
Adapt api-gateway-controller registry

triffer:
**Description**

- Change registry from `eu.gcr.io/kyma-project` to `europe-docker.pkg.dev/kyma-project/prod`

**Related issue(s)**


barchw:
@triffer I missed that in review, but we should also update the base images in Dockerfile*

Issue #313, opened 2023-04-13
[Istio JWT] requestPrincipal check in iss/* format


strekm:
**Description**

Following Istio docu  of Istio [AuthorizationPolicy](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source) field requestPrincipals should be in `<ISS>/<SUB>`. Atm this field is set to `*` and can be implemented in `<ISS>/*` as issuer is mandatory field. 

ACs:
- [x] requestPrincipal on Istio AuthorizationPolicy is set in `<ISS>/*` format

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] create release and bump in kyma~~

**Attachments**
https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source

**PR**
- [x] https://github.com/kyma-project/api-gateway/pull/319

PR #319, opened 2023-04-18
Add requestPrincipal with issuer

werdes72:
See: #313


internal/builders/istio.go
```diff
@@ -139,10 +140,25 @@ func (rf *FromBuilder) Get() *v1beta1.Rule_From {
...
 	// Only support one source at the moment
-	source := v1beta1.Source{RequestPrincipals: []string{"*"}}
+	var requestPrincipals []string
+	for _, strategy := range accessStrategies {
+		if strategy.Name == "jwt" {
+			authentications := &Authentications{
+				Authentications: []*Authentication{},
+			}
+			if strategy.Config != nil {
+				_ = json.Unmarshal(strategy.Config.Raw, authentications)
```
videlov, line 153: please do not ignore errors when unmarshalling Raw content. did you test if there is `Authorizations` in config as well? if not please add a unit test for it.

videlov (changes requested):

videlov:
please also go over DoDs for the task, e.g. docs and integration test

internal/builders/istio.go
werdes72: It's the same way of getting that issuer as in:
https://github.com/kyma-project/api-gateway/blob/main/internal/builders/istio.go#L332
We're not checking for error there either.

Added `Authorizations` checking in unit test.

videlov: we should fix both then? this is hiding a potential error, which relies solely on validation and we can not be certain all cases are checked. having error in such places will reduce time when investigating issues.

videlov: seems like `json.Umarshal` on validate is enough to caught issues on Raw config level.

videlov (approved):

Issue #278, opened 2023-03-29
Authorization Policy selector uses app name instead of service selectors


barchw:
**Description**
Authorization Policy is generated based on `APIRule` service, either `spec.service` or `spec.rules[*].service`. The current behaviour assumes that the deployment on which the Authorization Policy is applied to has label `app: ${service.name}`, but this might not always be the case. https://github.com/kyma-project/api-gateway/blob/261a486f70585107702ff6aaaeee57870c9c5efc/internal/processing/istio/authorization_policy_processor.go#L120

**Expected result**
Authorization Policy selector is generated based on actual in-cluster APIRule target service, falling back to name of the service if it is not present on cluster (or resulting in APIRule error)

**Actual result**
The AP selector is set to `app: ${service.name}`

**PR**
- [x] https://github.com/kyma-project/api-gateway/pull/338
- [x] https://github.com/kyma-project/kyma/pull/17447

PR #338, opened 2023-05-02
Fetching workload label selector from service on cluster directly

videlov:
**Description**

- Fetching workload label selector from service on cluster directly

**Related issue(s)**

https://github.com/kyma-project/api-gateway/issues/278


videlov:
> Do you think that adding a integration scenario for this makes sense?

yes ofc, why not? generally if it is possible to test something that has specifics while running on a real cluster with integration test I would do it. Even thought, I would also make sure that calling endpoint works as well, which I did not test. Perhaps I should extend the test.

barchw (changes requested):
The label selector needs to be applied both to Authorization Policy and Request Authentication

tests/integration/features/istio-jwt/istio_jwt.feature
```diff
@@ -128,3 +128,11 @@ Feature: Exposing endpoints with Istio JWT authorization strategy
     And JwtTokenFromParams: Calling the "/ip" endpoint with a valid "JWT" token from default header should result in status between 400 and 403
     And JwtTokenFromParams: Calling the "/ip" endpoint with a valid "JWT" token from parameter "jwt_token" should result in status between 200 and 299
     And JwtTokenFromParams: Teardown httpbin service
+
+  Scenario: Calling a helloworld endpoint with custom label selector service
+    Given CustomLabelSelector: There is a helloworld service with custom label selector name "custom-name"
+    And CustomLabelSelector: There is an endpoint secured with JWT on path "/ip"
+    When CustomLabelSelector: The APIRule is applied
+    Then CustomLabelSelector: Calling the "/ip" endpoint without a token should result in status between 400 and 403
+    And CustomLabelSelector: Calling the "/ip" endpoint with a valid "JWT" token should result in status between 200 and 299
```
barchw, line 137: Helloworld app doesn't have /ip endpoint
```suggestion
    And CustomLabelSelector: There is an endpoint secured with JWT on path "/hello"
    When CustomLabelSelector: The APIRule is applied
    Then CustomLabelSelector: Calling the "/hello" endpoint without a token should result in status between 400 and 403
    And CustomLabelSelector: Calling the "/hello" endpoint with a valid "JWT" token should result in status between 200 and 299
```

barchw (changes requested):

internal/processing/istio/authorization_policy_processor.go
```diff
@@ -12,10 +13,13 @@ import (
...
 	"istio.io/api/security/v1beta1"
 	securityv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	audienceKey string = "request.auth.claims[aud]"
+	audienceKey          string = "request.auth.claims[aud]"
+	appSelectorLabel     string = "app"
+	serviceSelectorLabel string = "service"
```
barchw, line 22: Both of those new variables are unused
```suggestion
```

barchw (changes requested):
Looks good either than the small linting issue, can be merged after we figure out how to propagate RBAC without updating Kyma resources

barchw (approved):

Issue #334, opened 2023-04-27
Istio JWT changes documentation


strekm:
**Description**

Create document describing differences between specification of ORY oathkeeper JWT and Istio JWT. Include changes if property is mandatory or not, format changes, functionality changes. Document needs to contain caution note that Istio JWT is not production ready.

**Reasons**

Describing changes in Istio based JWT handler.

**DoD:**
- [x] provide documentation

**PR:**
- [ ] https://github.com/kyma-project/api-gateway/pull/347

**Attachments**
part of: https://github.com/kyma-project/api-gateway/issues/30

PR #347, opened 2023-05-08
JWT access strategy docs for Ory and Istio comparison

videlov:
**Description**

- JWT access strategy documentation for Ory and Istio and comparison

**Related issue(s)**

https://github.com/kyma-project/api-gateway/issues/334


docs/jwt-ory-to-istio.md
```diff
@@ -0,0 +1,131 @@
...
+  rules:
+    - path: /.*
+      methods: ["GET"]
+      mutators: []
+      accessStrategies:
+        - handler: jwt
+          config:
+            authentications:
+            - issuer: $ISSUER
+              jwksUri: $JWKS_URI
```
triffer, line 74: In the Oathkeeper example we used `$JWKS_URI1`, maybe we should use the same name here.

docs/jwt-ory-to-istio.md
```diff
@@ -0,0 +1,131 @@
...
+              fromHeaders:
```
triffer, line 75: Why do we have the additional configuration like fromHeaders, fromParameters, requiredScopes or audiences only in the Istio and not in the Oathkeeper example? Should we have it this detailed here at all?

videlov: Yes, this makes sense will adapt it. Thx.

docs/jwt-ory-to-istio.md
```diff
@@ -0,0 +1,131 @@
...
+    port: 8080
+  rules:
+    - path: /.*
+      methods: ["GET"]
+      mutators: []
+      accessStrategies:
+        - handler: jwt
+          config:
+            authentications:
+            - issuer: $ISSUER
```
triffer, line 73: In the Oathkeeper example we used `$ISSUER1`, maybe we should use the same name here.
```suggestion
            - issuer: $ISSUER1
```

docs/jwt-ory-to-istio.md
```diff
@@ -0,0 +1,131 @@
...
+              jwksUri: $JWKS_URI
```
triffer, line 74: In the Oathkeeper example we used $JWKS_URI1, maybe we should use the same name here.
```suggestion
              jwksUri: $JWKS_URI1
```

nataliasitko (approved):

PR #351, opened 2023-05-10
Make authentications required

barchw:
**Description**

- Validate that authentications are populated for Istio JWT


videlov (approved):

videlov (changes requested):
Should we also adapt the json rules in here:
https://github.com/kyma-project/api-gateway/blob/main/api/v1beta1/apirule_types.go#L158

eventually adding kubebuilder `// +optional` for `authorizations` only

videlov (approved):

PR #402, opened 2023-06-22
Validate label selector only in Istio JWT

barchw:
**Description**

- Move label selector validation to Istio JWT

### DIFF
--- a/internal/validation/validate.go
+++ b/internal/validation/validate.go
@@ -3,6 +3,8 @@ package validation
 import (
 	"context"
 	"fmt"
+	"github.com/go-logr/logr"
+	"google.golang.org/appengine/log"
 	"strings"
 
 	"github.com/kyma-project/api-gateway/internal/helpers"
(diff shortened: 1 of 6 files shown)


internal/validation/validate.go
```diff
@@ -212,7 +204,7 @@ func (v *APIRuleValidator) validateRules(ctx context.Context, client client.Clie
 			if err != nil {
 				problems = append(problems, Failure{
 					AttributePath: attributePathWithRuleIndex + ".service",
-					Message:       "No label selectors found for service",
+					Message:       "Couldn't get label selectors for service",
```
videlov, line 207: should this stay? this is general validation on path level and not only for JWT case, you removed it in `validateService()` on line 174?

internal/validation/validate_test.go
```diff
@@ -953,44 +945,7 @@ var _ = Describe("Validate function", func() {
 		}).Validate(context.TODO(), fakeClient, input, networkingv1beta1.VirtualServiceList{})
 
 		//then
-		Expect(problems).To(HaveLen(1))
-		Expect(problems[0].AttributePath).To(Equal(".spec.service"))
-		Expect(problems[0].Message).To(Equal("No label selectors found for service"))
-	})
-
-	It("Should fail with service on path level without labels selector", func() {
```
videlov, line 961: you remove on path level validate test, that is why you did not cover case / usage of no label selector on a service used on path level.

internal/validation/validate.go
barchw: I would say that this can stay, as we expect that a service is present anyway

internal/validation/validate_test.go
barchw: I re-added the test https://github.com/kyma-project/api-gateway/pull/402/commits/b4f7dca6dc52fa9fdaf0bb43f39f2cacd7366db0

internal/validation/validate.go
barchw: I guess we may want to skip failing the API Rule generation maybe? WDYT

videlov: Would this work at runtime, if not, I would say we fail as early as possible.

barchw: This would work, as we can create e.g VirtualService, just the service won't be present (that was how it worked before)
It only matters for Istio JWT whether the Service is actually there for the APs creation

videlov: in this case I would be okay with just warning, as long as scenario works

barchw: https://github.com/kyma-project/api-gateway/pull/402/commits/080bcc10892851990cc32a139cd164bbbf9f0731

videlov (approved):

Issue #169, opened 2023-01-18
Provide support for configuring request timeout in APIRule


abbi-gaurav:
For long-running requests, the getway timout occurs. Currently, there is no way to control the timeout parameter via APIRule.

videlov:
As discussed, we would like to introduce it per service / API-Rule for most flexibility.

ptesny:
As it looks like certain target system may be slow to respond; at least initially; thus being able to add a request timeout to the virtual service of an API rule would make like of business developers way easier

ptesny:
my recent testing prove that the 180 seconds HTTP timeout on the virtual service side has no effect;

ptesny:
It seems that for the virtual service timeout to work the istio ingress gateway LB timeout needs to adjusted accordingly, for instance:
AWS:
```
$ kubectl annotate svc -n istio-system istio-ingressgateway service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout="3600" 
service/istio-ingressgateway annotated
```
AZURE:
```
$ kubectl annotate svc -n istio-system istio-ingressgateway service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout="30" 
service/istio-ingressgateway annotated
```

triffer:
We investigated further where timeouts can occur and how we can change this configuration.
There are different levels where a timeout can occur.

### Timeout in Virtual Service
This is set by the api-gateway to 180s for all virtual services.

As [already discussed](https://github.com/kyma-project/api-gateway/issues/169#issuecomment-1387192208), it would be useful if this can be configured in the APIRule.

### Timeout caused by Ory Oathkeeper
If the workload is secured by Oathkeeper, e.g. by using the noop, oauth2 or jwt access strategy, there is a default timeout of 120s. Once the 120s is reached, Oathkeeper considers this request as failed. Istio then retries the request and sends it to Oathkeeper again until the default value of two retries or the timeout configured in the virtual service is reached.

It is not possible to disable the timeout in Oathkeeper, so it would be best to set this timeout to a very high value such as 2 hours, as the virtual service will ensure that the timeout is handled correctly.

### Timeout in Hyperscaler-specific load balancers
We checked the behaviour for long-running requests for AWS, GCP and Azure. For GCP and Azure, we were able to successfully run a one-hour request after exposing the workload with a virtual service with a configured timeout of 2 hours. 

[...]

PR #442, opened 2023-08-02
Provide support for configuring request timeout in APIRule

triffer:
**Description**

Instead of using the default timeout of 180s, this PR adds support for an optional timeout configuration on the spec and rule level. If there is no timeout defined, the default of 180s is used. The timeout can be configured in seconds instead of duration, because Istios virtual service doesn't support milliseconds and an [issue for duration](https://github.com/kubernetes/apiextensions-apiserver/issues/56) exists, that would require additional validation.

- Add timeout configuration on spec and rule level. The maximum timeout can be 3900s (65m), because of hyperscaler restrictions.

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/169


controllers/api_controller_integration_test.go
```diff
@@ -174,7 +174,7 @@ var _ = Describe("APIRule Controller", Serial, func() {
 									Route(builders.RouteDestination().Host(testOathkeeperSvcURL).Port(testOathkeeperPort)).
 									Headers(builders.NewHttpRouteHeadersBuilder().SetHostHeader(serviceHost).Get()).
 									CorsPolicy(defaultCorsPolicy).
-									Timeout(time.Second * helpers.DEFAULT_HTTP_TIMEOUT))
+									Timeout(time.Second * 180))
```
barchw, line 177: This `180` value is shared across every test, maybe we should have it as constant at least here?

api/v1beta1/apirule_types.go
```diff
@@ -181,3 +185,8 @@ type JwtHeader struct {
 	// +optional
 	Prefix string `json:"prefix,omitempty"`
 }
+
+// Timeout for HTTP requests in seconds. The timeout can be configured up to 3900 seconds (65 minutes).
+// +kubebuilder:validation:Minimum=1
+// +kubebuilder:validation:Maximum=3900
+type Timeout uint16 // We use unit16 instead of a time.Duration because there is a bug with duration that requires additional validation of the format. Issue: checking https://github.com/kubernetes/apiextensions-apiserver/issues/56
```
barchw, line 192: We could probably configure it as it was done [here](https://github.com/openshift/hive/pull/1690/files)
Do you think that this is not the right approach?

I guess limiting the maximum value would be harder probably

internal/processing/processors/virtual_service_processor_test.go
```diff
@@ -78,6 +84,65 @@ var _ = Describe("Virtual Service Processor", func() {
...
+	It("should return default when no timeout is set", func() {
+		// given
+		apiRuleSpec := gatewayv1beta1.APIRuleSpec{}
+		rule := gatewayv1beta1.Rule{}
+
+		// when
+		timeout := processors.GetVirtualServiceHttpTimeout(apiRuleSpec, rule)
+
+		// then
+		Expect(timeout).To(Equal(time.Second * 180))
```
barchw, line 97: The 180 default is not really known from the test's perspective.
Maybe adjust the description to `should return default 180s when no timeout is set`?

Or we should keep the timeout in helpers package and share it across all files that use it

barchw (changes requested):

api/v1beta1/apirule_types.go
triffer: I actually had this as a first idea and implementation. But the max value would need an implementation in the validation. Also having the pattern validation is not so nice, since we'd only allow s, m and h. When the user adds a configuration that will not match the pattern, a regex is returned and this is not a good UX imho.

That's why considering all of this I decided to go for the simple and straight forward solution.

controllers/api_controller_integration_test.go
triffer: Good idea, just did a find replace and didn't check that it can be an constant in the test.

internal/processing/processors/virtual_service_processor_test.go
triffer: I'd not keep it in the helpers package, since it's only relevant for the virtual service. So the reason why I moved it closer to the virtual service is cohesion and slowly removing the helper package.

internal/processing/processors/virtual_service_processor.go
```diff
@@ -62,3 +65,15 @@ func (r VirtualServiceProcessor) getObjectChanges(desiredVs *networkingv1beta1.V
 		return processing.NewObjectCreateAction(desiredVs)
 	}
 }
+
+func GetVirtualServiceHttpTimeout(apiRuleSpec gatewayv1beta1.APIRuleSpec, rule gatewayv1beta1.Rule) time.Duration {
+	if rule.Timeout != nil {
+		return time.Duration(*rule.Timeout) * time.Second
```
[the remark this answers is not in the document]
triffer: Oh, I totally agree 😄

internal/processing/processors/virtual_service_processor_test.go
triffer: I'm not a big fan of re-using constant from real code in tests or exporting them for test only, since changing this values often is a break of the functional contract of your code. If we change the value it should break some tests, to make it obvious that some external behaviour changed. If it's exported it can be easily overseen during refactoring.

docs/api-rule-cr.md
```diff
@@ -76,23 +77,25 @@ spec:
...
+| **spec.rules.mutators**          |  **NO**   | Specifies the list of [Oathkeeper](https://www.ory.sh/docs/next/oathkeeper/pipeline/mutator) or Istio mutators.                                                                                                                                                                                        |
+| **spec.rules.accessStrategies**  |  **YES**  | Specifies the list of access strategies. Supported are [Oathkeeper](https://www.ory.sh/docs/next/oathkeeper/pipeline/authn) `oauth2_introspection`, `jwt`, `noop` and `allow`. We also support `jwt` as [Istio](https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/) access strategy. |
+| **spec.rules.timeout**           |  **NO**   | Specifies the timeout for HTTP requests for the rule in seconds. The maximum timeout is limited to 3900 seconds (65 minutes). Timeout definitions at this level have a higher precedence than the timeout definition at the **spec.timeout** level.                                                    |
```
nataliasitko, line 98:
```suggestion
| **spec.rules.timeout**           |  **NO**   | Specifies the timeout, in seconds, for HTTP requests made to the Rule. The maximum timeout is limited to 3900 seconds (65 minutes). Timeout definitions set at this level take precedence over any timeout defined at the **spec.timeout** level.                                                   |
```
Should we also mention here that if the value is not specified, the default timeout of 180 seconds is used?

nataliasitko: > I also thought about this, but if no value is defined, spec.timeout kicks in and if there nothing is defined, we have the docs about the default. So I thought this "layered" doc approach would be better, to not confuse users.

Yes, let's not mention it here when we have this information for the **spec.timeout** field. I didn't notice that.

> Also I think `Specifies the timeout, in seconds, for HTTP requests made to the Rule.` is not correct and `Specifies the timeout, in seconds, for HTTP requests made to **spec.rules.path**.` is more accurate. What do you think?

Sure, sounds good to me :) Thanks for spotting that

triffer: 👍 Thanks, then I'll update this
```suggestion
| **spec.rules.timeout**           |  **NO**   | Specifies the timeout, in seconds, for HTTP requests made to **spec.rules.path**. The maximum timeout is limited to 3900 seconds (65 minutes). Timeout definitions set at this level take precedence over any timeout defined at the **spec.timeout** level.                                                    |
```

barchw:
Httpbin has an endpoint `/delay/{delay}` that will respond to a request with a delay specified in the path parameter (max 10s). Should we create a integration test that uses that functionality?

barchw (approved):

nataliasitko (approved):

PR #481, opened 2023-08-25
Create API-Gateway-CRD-proposal.md

barchw:
**Description**

- Add CRD proposal


docs/API-Gateway-CRD-proposal.md
```diff
@@ -0,0 +1,146 @@
...
+  #        memory: 128Mi
+  
+  # Kyma gateway should always be available
+  #disableKymaGateway: false
+  gateways:
+    - namespace: "some-ns" # Required
+      name: "gateway1" # Required
+      servers:
+        - hosts: # Creating  more than one for the same host:port configuration should result in  Warning
+            - dnsName: "goat.example.com"
```
triffer, line 50: I think "dnsName" can lead to confusion. Also we are using host in the APIRule. Since those two belong together, I think it makes sense to not have them different named. Maybe "hostname" or "host" is a better name? In K8s Gateway API "hostname" is used.

docs/API-Gateway-CRD-proposal.md
```diff
@@ -0,0 +1,146 @@
...
+              certificate: "goat-certificate" # If not defined, generate Gardener certificate
+            - dnsName: "goat1.example.com"
+              DNSProviderSecret: "my-namespace/dns-secret" # If provided generate a DNS Entry with Gardener
```
triffer, line 53: Is there a reason why this is capitalized?

docs/API-Gateway-CRD-proposal.md
```diff
@@ -0,0 +1,146 @@
...
+              default: true # Use as deafult domain for API Rules
+            - dnsName: "goat1.example.com"
+              DNSProviderSecret: "my-namespace/dns-secret" # If provided generate a DNS Entry with Gardener 
+          port:
+            number: 80
+            name: http
+            protocol: HTTP
+          httpsRedirect: true # If on Protocol = HTTPS, set Warning
+        # We should consider configuration for MTLS gateway
+        # TLS: mutual might be considered in
```
triffer, line 69: What do we do about the tls config?

triffer (commented):
I like the example use cases 👍

docs/API-Gateway-CRD-proposal.md
barchw: The name is from gardener: https://github.com/gardener/external-dns-management/blob/846d76ec5a07ab2b904645bb115bc9ffe5ce239b/pkg/apis/dns/v1alpha1/dnsentry.go#L63C7-L63C7

docs/API-Gateway-CRD-proposal.md
```diff
@@ -0,0 +1,105 @@
...
+
+This document describes proposed API for installing APIGateway component.
+
+## Proposed CR structure
+
+```yaml
+kind: APIGateway
+spec:
+  enableKymaGateway: true # Part of the custom resource gby default (or true if not defined)
+  defaultHost: "example.com" # Use as deafult host for API Rules, if not defined use Kyma host if `enableKymaGateway: true`, if both are false, require full host in API Gateway
```
triffer, line 11:
```suggestion
  defaultHost: "example.com" # Use as default host for API Rules, if not defined use Kyma host if `enableKymaGateway: true`, if both are false, require full host in API Gateway
```

triffer (dismissed):

triffer (dismissed):

triffer (approved):

nataliasitko (approved):

PR #515, opened 2023-09-13
Discover default domain from gateway

barchw:
**Description**

- Discover host from default gateway

**Related issue(s)**


internal/helpers/default_domain.go
```diff
@@ -1,10 +1,19 @@
 package helpers
```
triffer, line 1: I think we should not build on the pattern to have generic packages like "helpers", "common", ... but try to find a good scope for new code and also move old code out of this packages.

internal/helpers/default_domain.go
```diff
@@ -23,3 +32,21 @@ func GetHostWithDefaultDomain(host, defaultDomainName string) string {
...
 }
+
+func GetDefaultDomainFromKymaGateway(ctx context.Context, k8sClient client.Client) (string, error) {
+	var gateway networkingv1beta1.Gateway
+	err := k8sClient.Get(ctx, types.NamespacedName{Namespace: gatewayNamespace, Name: gatewayName}, &gateway)
+	if err != nil {
+		return "", err
+	}
+
+	if len(gateway.Spec.GetServers()) != 2 {
```
triffer, line 43: I think we are making a lot of assumptions about the structure of the gateway. Even though this resource is under our control, for some reason it can be changed from being http the first entry. I think it would be more resilient and maybe easier to understand what data we want, if we iterate the servers and filter for the https server and using the host of this.

internal/helpers/default_domain_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+		// when
+		host, err := GetDefaultDomainFromKymaGateway(context.Background(), client)
+
+		// then
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(host).To(Equal("local.kyma.dev"))
+	})
+
+	It("should return error if gateway does not have to servers", func() {
```
triffer, line 88: If this implementation will stay, then we should fix this typo.
```suggestion
	It("should return error if gateway does not have two servers", func() {
```

triffer (commented):
We should add some documentation about the current behaviour of default domain and when it's reconciled in the technical design documentation of the APIRule controller. I think we can wait for the user documentation until we have it available in the Gateway CR. 
We might also want to update the `operator-contributor-skr-overview.svg` to contain this change.

controllers/gateway/api_controller_integration_test.go
```diff
@@ -43,6 +44,127 @@ var _ = Describe("APIRule Controller", Serial, func() {
...
+
+			// when
+			Expect(c.Create(context.TODO(), svc)).Should(Succeed())
+			Expect(c.Create(context.TODO(), apiRule)).Should(Succeed())
+			defer func() {
+				apiRuleTeardown(apiRule)
+				serviceTeardown(svc)
+			}()
+
+			expectApiRuleStatus(apiRuleName, gatewayv1beta1.StatusError)
```
triffer, line 71: Shouldn't the test end here? I was confused that there is an APIRule update after that, since the test case mentions the error as the verification step.Do we have details in the error description that can be user for further validation, so we can remove the rest of the code in this test?

controllers/gateway/api_controller_integration_test.go
```diff
@@ -43,6 +44,127 @@ var _ = Describe("APIRule Controller", Serial, func() {
 		defaultHttpTimeout         = time.Second * 180
 	)
 
+	Context("check default domain logic", func() {
+
+		It("should have an error when creating an APIRule without a domain in cluster without kyma-gateway", func() {
+			updateJwtHandlerTo(helpers.JWT_HANDLER_ORY)
```
triffer, line 50: I think it would be better to use Istio instead of Ory, if we want to have a test only for one of this handlers, since Istio is the handler that will stay.
If we drop the Ory handler we'd have to adapt this test for Istio and in worst-case it will get deleted, because it's an Ory test.

docs/contributor/04-10-technical-design.md
```diff
@@ -24,6 +24,8 @@ The API Rule Controller is a [Kubernetes controller](https://kubernetes.io/docs/
 The controller is responsible for handling the [APIRule CR](../user/custom-resources/apirule/01-40-apirule-custom-resource.md).  
 Additionally, the controller watches the [`api-gateway-config`](../user/custom-resources/apirule/01-40-apirule-custom-resource.md#jwt-access-strategy) to configure the JWT handler.
 
+The API Rule Controller has a conditional dependency to API Gateway controller in terms of default API Rule domain. If no domain is configured in API Gateway CR (currently not yet possible), API Rule controller will use default kyma gateway domain as the default value for created Virtual Services
```
triffer, line 27:
```suggestion
The API Rule Controller has a conditional dependency to API Gateway controller in terms of default API Rule domain. If no domain is configured in API Gateway CR (currently not yet possible), API Rule controller will use default kyma gateway domain as the default value for created Virtual Services.
```

controllers/gateway/api_controller_integration_test.go
barchw: I wanted to additionally test that supplying a full domain `httpbin.local.kyma.dev` will fix the error on APIRule
If you feel like it would be better to remove that, I can do it

triffer: In my opinion if we want to do it like that, then we should adjust the test description, so it's clear that the error is not the final state of the test.

barchw: 15a0087...56d2aaa

triffer (dismissed):

docs/contributor/04-10-technical-design.md
```diff
@@ -24,6 +24,8 @@ The API Rule Controller is a [Kubernetes controller](https://kubernetes.io/docs/
 The controller is responsible for handling the [APIRule CR](../user/custom-resources/apirule/01-40-apirule-custom-resource.md).  
 Additionally, the controller watches the [`api-gateway-config`](../user/custom-resources/apirule/01-40-apirule-custom-resource.md#jwt-access-strategy) to configure the JWT handler.
 
+The API Rule Controller has a conditional dependency to API Gateway controller in terms of default API Rule domain. If no domain is configured in API Gateway CR (currently not yet possible), API Rule controller will use default kyma gateway domain as the default value for created Virtual Services.
```
nataliasitko, line 27:
```suggestion
API Rule Controller has a conditional dependency to API Gateway Controller in terms of the default APIRule domain. If no domain is configured in APIGateway CR, API Rule Controller uses the default Kyma Gateway domain for creating Virtual Services.
```

docs/contributor/04-10-technical-design.md
```diff
@@ -24,6 +24,8 @@ The API Rule Controller is a [Kubernetes controller](https://kubernetes.io/docs/
 The controller is responsible for handling the [APIRule CR](../user/custom-resources/apirule/01-40-apirule-custom-resource.md).  
 Additionally, the controller watches the [`api-gateway-config`](../user/custom-resources/apirule/01-40-apirule-custom-resource.md#jwt-access-strategy) to configure the JWT handler.
 
+API Rule Controller has a conditional dependency to API Gateway Controller in terms of the default APIRule domain. If no domain is configured in APIGateway CR, API Rule Controller uses the default Kyma Gateway domain as the default value for creating Virtual Services.
+
```
nataliasitko, line 28:
```suggestion
For now, you must configure a domain in APIGateway CR. The possibility of skipping this step and using the default domain will be available at a later time. See the [epic task](https://github.com/kyma-project/api-gateway/issues/130).
```

nataliasitko (approved):

PR #539, opened 2023-09-18
Empty/stub release workflow for testing purposes

jaroslaw-pieszka:
**Description**

Create release workflow accepting proper parameters and creating only lightweight tag.
It is supposed to allow testing workflow implemented on the `mod-dev` branch.

**Related issue(s)**

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -11,3 +11,14 @@ on:
         description: "Changelog since"
         default: ""
         required: false
+
+jobs:
+  create-draft:
+    name: Create draft release
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Create lightweight tag
+        run: |
+          git tag ${{ github.event.inputs.name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,24 @@
...
+        default: ""
+        required: false
+
+jobs:
+  create-draft:
+    name: Create draft release
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Create lightweight tag
```
kolodziejczak, line 21: Shouldn't we checkout before operating on the repository?

kolodziejczak (changes requested):

.github/workflows/create-release.yaml
jaroslaw-pieszka: I am not absolutely sure it is necessary, so I added checkout.

kolodziejczak (approved):

Issue #559, opened 2023-09-25
APIGateway CR deletion support


strekm:
**Description**

API Gateway controller should react on APIGateway CR deletion event. If there are no VS/APIRule referencing default gateway, controller should remove default gateway. In case there are any Istio VS/APIRule referencing default gateway these resources should be listed in logs and APIGateway CR should be in warning status indicating user action. Once user will delete or modify logged resources default gateway can be removed safely and deletion of APIGateway CR can continue.

ACs:
- [x] user can delete APIGateway CR
- [x] APIGateway CR is in warning state if there are any resources using default gateway
- [x] resources blocking deletion listed in logs
- [x] default gateway removed
- [x] adapted deletion conditions:
* Kyma Gateway is blocked by referring APIRules and VirtualServices
* API-Gateway CR is blocked by any APIRules on the cluster
- [x] [godog integration test/scenario/feature](https://github.com/kyma-project/api-gateway/pull/567/commits/9a7d8e3de109f87f7dac1f5e9f1d8f80121c8ce9) implemented

**Reasons**

api-gateway modularisation

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] create release and bump in kyma~~
- [x] PR reviewer will verify code coverage and evaluate if it is acceptable

**Attachments**
part of: https://github.com/kyma-project/api-gateway/issues/130

**PRs**
- [x] https://github.com/kyma-project/api-gateway/pull/567

PR #537, opened 2023-09-18
Add Kyma Gateway

triffer:
**Description**

Changes proposed in this pull request:

- Extend APIGateway CR to enable and disable Kyma Gateway
- Creation and deletion of Kyma Gateway with different handling for Gardener and other environments

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/532

PR #567, opened 2023-09-26
API-Gateway module CR blocking deletion

videlov:
**Description**

- API-Gateway module CR blocking deletion

**Related issue(s)**

https://github.com/kyma-project/api-gateway/issues/559


internal/operator/reconciliations/api_gateway/reconciliation.go
```diff
@@ -0,0 +1,149 @@
...
+	Reconcile(ctx context.Context, apiGatewayCR operatorv1alpha1.APIGateway, apiGatewayResourceListPath string) (operatorv1alpha1.APIGateway, described_errors.DescribedError)
+}
+
+type Reconciliation struct {
+	Client client.Client
+}
+
+const (
+	reconciliationFinalizer string = "apigateways.operator.kyma-project.io/api-gateway-reconciliation"
+	dafaultGatewayName      string = "kyma-gateway"
```
triffer, line 31: There is a typo `dafault` . Also I think we should expose this variables from the gateway package from PR https://github.com/kyma-project/api-gateway/pull/537 and use it here.

internal/operator/reconciliations/api_gateway/reconciliation.go
```diff
@@ -0,0 +1,149 @@
...
+			for _, res := range clientResources {
+				ctrl.Log.Info("Customer resource is blocking API-Gateway deletion", res.GVK.Kind, fmt.Sprintf("%s/%s", res.Namespace, res.Name))
+			}
+
+			return apiGatewayCR, described_errors.NewDescribedError(fmt.Errorf("could not delete API-Gateway module instance since there are %d custom resource(s) present that block its deletion", len(clientResources)),
+				"There are custom resource(s) that block the deletion. Please take a look at kyma-system/api-gateway-controller-manager logs to see more information about the warning").DisableErrorWrap().SetWarning()
+		}
+
+		//Temporarily having this simple solution for deletion of Kyma default gateway before we migrate whole management of Kyma default gateway from Istio component
+		if err := deleteDefaultGateway(ctx, i.Client); err != nil {
```
triffer, line 88: We already have some logic for this in PR https://github.com/kyma-project/api-gateway/pull/537. We should not have it duplicated.

internal/operator/reconciliations/api_gateway/reconciliation.go
videlov: Good catch, yes, it would be best to have it in gateway package and reuse it from there.

internal/operator/reconciliations/api_gateway/reconciliation.go
videlov: Yes, as I did add comment above just temporarily for these PR, I somehow missed the other PR and the activity on this resource handling. We should adopt functionality from gateway package.

internal/operator/reconciliations/api_gateway/reconciliation.go
```diff
@@ -0,0 +1,149 @@
...
+	Client client.Client
+}
+
+const (
+	reconciliationFinalizer string = "apigateways.operator.kyma-project.io/api-gateway-reconciliation"
+	dafaultGatewayName      string = "kyma-gateway"
+	dafaultGatewayNS        string = "kyma-system"
+)
+
+var checkDefaultGatewayReference = func(ctx context.Context, c client.Client, res resources.Resource) bool {
```
triffer, line 36: I think this logic should also be moved to the gateway package, since this is tightly coupled to it. Also we can re-use this logic here to block the deletion of the Kyma Gateway, because right now it [blocks on any APIRule](https://github.com/kyma-project/api-gateway/blob/mod-dev/internal/gateway/kyma_gateway.go#L30). Having different criteria doesn't make sense.

internal/operator/reconciliations/api_gateway/reconciliation.go
```diff
@@ -0,0 +1,149 @@
...
+			for _, res := range clientResources {
+				ctrl.Log.Info("Customer resource is blocking API-Gateway deletion", res.GVK.Kind, fmt.Sprintf("%s/%s", res.Namespace, res.Name))
+			}
+
+			return apiGatewayCR, controllers.WarningStatus(fmt.Errorf("could not delete API-Gateway module instance since there are %d custom resource(s) present that block its deletion", len(clientResources)),
+				"There are custom resource(s) that block the deletion. Please take a look at kyma-system/api-gateway-controller-manager logs to see more information about the warning")
+		}
+
+		//Temporarily having this simple solution for deletion of Kyma default gateway before we migrate whole management of Kyma default gateway from Istio component
+		if err := deleteDefaultGateway(ctx, i.Client); err != nil {
```
triffer, line 89: I think we still need to move this to the gateway package. All logic about the Kyma gateway should be encapsulated there. We can expose a function to call it here. Also there is a Kyma Gateway finalizer that will not be handled correctly if deleted like this.

internal/operator/reconciliations/api_gateway/reconciliation.go
videlov: Will move it to gateway package, but I am unsure about the logic - https://github.com/kyma-project/api-gateway/issues/559 states it should check if APIRule is referring the gateway. On the mentioned code, it blocks on any APIRule. If I am remembering correctly first we agreed not to check if APIRule refers the gateway, but thought this changed. Should we sync on this again?

videlov: Decided on that, we move the condition for referring APIRules and VSs to gateway deletion and check for existing any APIRules to API-Gateway CR deletion conditions.

docs/contributor/04-10-technical-design.md
```diff
@@ -3,25 +3,27 @@
...
...teway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md), the controller will check before deleting it if there any APIRule or Istio [VirtualService](https://istio.io/latest/docs/reference/config/networking/virtual-service) resources which are referring to the Kyma default [Gateway](https://istio.io/latest/docs/reference/config/networking/gateway/) (`kyma-system/kyma-gateway`). If there are any found, they will be listed in the logs of the controller and [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md)'s status will be set to `Warning` explaining that there are resources blocking the deletion. Deletion of the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md) is protected with a finalizer named `gateways.operator.kyma-project.io/api-gateway-reconciliation`. Once there are no more APIRule and VirtualService resources blocking the deletion, the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md) will be deleted. Deletion of [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md) also deletes Kyma default Gateway.
```
nataliasitko, line 21:
```suggestion
Before deleting the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md), the controller first checks if there are any APIRule or [Istio VirtualService](https://istio.io/latest/docs/reference/config/networking/virtual-service) resources that reference the default Kyma [Gateway](https://istio.io/latest/docs/reference/config/networking/gateway/) `kyma-system/kyma-gateway`. If any such resources are found, they are listed in the logs of the controller, and the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md)'s status is set to `Warning` to indicate that there are resources blocking the deletion. 
The `gateways.operator.kyma-project.io/api-gateway-reconciliation` finalizer protects the deletion of the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md). Once no more APIRule and VirtualService resources are blocking the deletion of the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md), the APIGateway CR can be deleted. Deleting the [APIGateway CR](../user/custom-resources/apigateway/04-00-apigateway-custom-resource.md) also deletes the default Kyma Gateway. 
```

internal/resources/resources.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return &ResourcesFinder{
+		ctx:           ctx,
+		logger:        logger,
+		client:        client,
+		configuration: finderConfiguration,
+	}, nil
+}
+
+func (i *ResourcesFinder) FindUserCreatedResources(checkIf resourceCondition) ([]Resource, error) {
```
triffer, line 83: Some documentation, especially how the provided function is used might be helpful. `checkIf` seems also very generic, maybe we can find a better name or make it a "boolean name" like `isResourceRelevant`.

internal/resources/resources_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+var sc *runtime.Scheme
+
+var _ = Describe("Resources", func() {
+	sc = runtime.NewScheme()
+	Expect(v1beta1.AddToScheme(sc)).To(Succeed())
+	Expect(networkingv1alpha3.AddToScheme(sc)).To(Succeed())
+	Expect(networkingv1beta1.AddToScheme(sc)).To(Succeed())
+
+	DescribeTable("FindUserCreatedIstioResourcesDescribe", func(ctx context.Context, logger logr.Logger, c client.Client, configuration resourceFinderConfiguration, conditionResult bool, want []Resource, wantErr bool) {
```
triffer, line 28: I think this should be renamed.
```suggestion
	DescribeTable("FindUserCreatedResourcesDescribe", func(ctx context.Context, logger logr.Logger, c client.Client, configuration resourceFinderConfiguration, conditionResult bool, want []Resource, wantErr bool) {
```

internal/tests/ginkgo_reporter.go
triffer, line 1: This code already exists in [api-gateway/tests/ginkgo_reporter.go](https://github.com/kyma-project/api-gateway/blob/mod-dev/tests/ginkgo_reporter.go).

tests/integration/jwt_handler.go
```diff
@@ -1,4 +1,4 @@
-package api_gateway
+package custom_resource
```
triffer, line 1: I think this was done accidentally.

tests/integration/main_test.go
```diff
@@ -1,4 +1,4 @@
-package api_gateway
+package custom_resource
```
triffer, line 1: I think this was done accidentally.

tests/integration/test_report.go
```diff
@@ -1,4 +1,4 @@
-package api_gateway
+package custom_resource
```
triffer, line 1: I think this was done accidentally.

tests/integration/testsuites/custom-domain/testsuite.go
```diff
@@ -108,15 +109,14 @@ func (t *testsuite) TearDown() {
 }
 
 func (t *testsuite) BeforeSuiteHooks() []func() error {
-	return []func() error{hooks.ApplyAndVerifyApiGatewayCrSuiteHook}
+	return []func() error{hooks.DeleteBlockingResourcesSuiteHook, hooks.ApplyAndVerifyApiGatewayCrSuiteHook}
```
triffer, line 112: Why is this added as a before hook? Is it a safe-guard to retry the deletion, if it failed in the After hook?
I think we should fail all the tests if a hook wasn't able to successfully clear the state, because all following tests might be affected.

tests/integration/pkg/hooks/hooks.go
```diff
@@ -215,3 +226,57 @@ func createApiGatewayCRObjectFromTemplate(name string) (v1alpha1.APIGateway, err
 	apiGateway.Name = name
 	return apiGateway, nil
 }
+
+func deleteBlockingResources(ctx context.Context) error {
```
triffer, line 230: I think a more generic approach, like it's in Istio would be better here. We add resources to a list if they are created during a test and they'll be deleted in the after hook.
I think we didn't implement it that way to keep it simpler for now and adapt it when we rewrite this integration tests?

controllers/operator/apigateway_controller.go
```diff
@@ -73,17 +82,21 @@ func (r *APIGatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
-	if kymaGatewayStatus := gateway.ReconcileKymaGateway(ctx, r.Client, &apiGatewayCR); !kymaGatewayStatus.IsReady() {
+	if kymaGatewayStatus := gateway.ReconcileKymaGateway(ctx, r.Client, &apiGatewayCR, APIGatewayResourceListDefaultPath); kymaGatewayStatus.IsError() || kymaGatewayStatus.IsWarning() {
 		return r.requeueReconciliation(ctx, apiGatewayCR, kymaGatewayStatus)
 	}
 
+	if apiGatewayStatus := r.apiGatewayReconciliation.Reconcile(ctx, &apiGatewayCR); apiGatewayStatus.IsError() || apiGatewayStatus.IsWarning() {
```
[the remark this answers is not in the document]
videlov: I agree, just am not sure if this was breaking some test. If not mistaken the order has to be exactly like that. But let see, I will move it up and we can check the tests.

controllers/operator/suite_test.go
```diff
@@ -55,8 +59,29 @@ import (
...
-	testNamespace     = "kyma-system"
+	testNamespace    = "kyma-system"
+	apiGatewayCRName = "default"
+
 	eventuallyTimeout = time.Second * 10
+
+	kymaNamespace   = "kyma-system"
+	kymaGatewayName = "kyma-gateway"
+
+	controlledList = `
```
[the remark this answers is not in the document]
videlov: Yes, I do not like this. It is only for the tests. I can not make it work for the `testenv` tests since they are running in some virtual KCP env and I did not find a way to provide the file available for this controller container env. If you know how, I would be happy to get rid of that.

internal/reconciliations/custom_resource/custom_resource.go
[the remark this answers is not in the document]
videlov: Yes, I agree. It is not really reconciling doing something substantial. Lets try to integrate it into the controller directly without reconciliation separation. Will refactor it

internal/reconciliations/gateway/kyma_gateway.go
```diff
@@ -16,32 +18,69 @@ const (
...
+		if err != nil {
+			return controllers.ErrorStatus(err, "Could not get customer resources from the cluster")
 		}
 
-		if !hasKymaGatewayFinalizer(*apiGatewayCR) {
-			ctrl.Log.Info("Kyma Gateway is disabled and no finalizer exists, reconciliation is skipped.")
-			return controllers.ReadyStatus()
+		if len(clientResources) > 0 {
+			for _, res := range clientResources {
+				ctrl.Log.Info("Custom resource is blocking Kyma Gateway deletion", res.GVK.String(), fmt.Sprintf("%s/%s", res.Namespace, res.Name))
```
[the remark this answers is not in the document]
videlov: why? how do you like it better?

internal/reconciliations/gateway/kyma_gateway_test.go
```diff
@@ -199,151 +226,123 @@ var _ = Describe("Kyma Gateway reconciliation", func() {
...
-		err := k8sClient.Get(context.Background(), client.ObjectKey{Name: kymaGatewayName, Namespace: kymaGatewayNamespace}, &kymaGateway)
-		Expect(err).ShouldNot(HaveOccurred())
+			Expect(status.NestedError().Error()).To(Equal("could not delete Kyma Gateway since there are 1 custom resource(s) present that block its deletion"))
+			Expect(status.Description()).To(Equal("There are custom resources that block the deletion. Please take a look at kyma-system/api-gateway-controller-manager logs to see more information about the warning"))
+		})
 	})
-
 })
 
-func testShouldDeleteKymaGatewayNonGardenerResources(updateApiGateway func(gw v1alpha1.APIGateway) v1alpha1.APIGateway) {
```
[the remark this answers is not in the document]
videlov: Yes, I agree but in this case it solves quite a code duplication and adding more cases avoids also adding more code duplication. Let me check if it could be better with less boolean-controlled logic and less code duplication.

internal/resources/resources.go
videlov: The goal of it was to be generic condition. I do not care much for names, will rename it to `isResourceRelevant`. If you ask me to add comments to the code, I politely refuse 😜

internal/reconciliations/gateway/dns_entry.go
```diff
@@ -77,10 +78,9 @@ func deleteDnsEntry(ctx context.Context, k8sClient client.Client, name, namespac
 }
 
 func fetchIstioIngressGatewayIp(ctx context.Context, k8sClient client.Client) (string, error) {
-
 	istioIngressGatewayNamespaceName := types.NamespacedName{
-		Name:      "istio-ingressgateway",
-		Namespace: "istio-system",
+		Name:      KymaGatewayName,
```
triffer, line 82: This needs to be reverted since it is not correct and will fail the reconciliation on Gardener clusters.

internal/reconciliations/gateway/dns_entry_test.go
```diff
@@ -84,8 +85,8 @@ var _ = Describe("DNSEntry", func() {
 func getTestIstioIngressGatewayDnsBasedService() corev1.Service {
 	return corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "istio-ingressgateway",
-			Namespace: "istio-system",
+			Name:      KymaGatewayName,
```
triffer, line 88: Needs to be updated as well, that's why I personally consider re-using implementation vars for tests a bad practice.

tests/integration/testsuites/custom-domain/testsuite.go
videlov: It is just to make sure we start clean. Perhaps we do not need it. Will try to have it just before deleting CR so I do not block because some resources are left uncleaned. It is a leftover of making scenarios work... faster :)

tests/integration/pkg/hooks/hooks.go
videlov: Yes, agree, I like the controlled approach in Istio. This is done just for the sake of moving the urgent API-Gateway impl tasks faster for the time being. We should invest more time into refactoring / making better the integration tests as a follow-up.

internal/reconciliations/gateway/kyma_gateway_test.go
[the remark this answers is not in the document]
triffer: I always consider readability more important than code duplication (especially in tests), since there will always be overlapping assertion scenarios. I have seen tests that called multiple functions for asserting resources. These tests had no code duplication, but it was a mess to understand all assertions are done for one specific test.

If it's too much effort, we don't have to do it now.

internal/reconciliations/gateway/kyma_gateway_test.go
```diff
@@ -199,151 +226,123 @@ var _ = Describe("Kyma Gateway reconciliation", func() {
...
+func testShouldDeleteKymaGatewayResources(updateApiGateway func(gw v1alpha1.APIGateway) v1alpha1.APIGateway, gardener bool, state controllers.State, nfMatcher types.GomegaMatcher, fMatcher types.GomegaMatcher, objs ...client.Object) controllers.Status {
```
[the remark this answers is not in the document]
triffer: Ok, but we are calling `testShouldDeleteKymaGatewayResources`
in the scope of `Should not delete Kyma Gateway [...]` test, that was confusing me. The function names suggests that it verifies that Kyma Gateway resources should be deletes, but the test names wants to verify the oppposite.

controllers/operator/apigateway_controller.go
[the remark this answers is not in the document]
videlov: Now I remember. If I move the more general condition first, I can not test for the more specific condition with the integration tests (because more general will fail first). I decided I like to have the tests for the more specific conditions instead of fail earlier. If you think failing early on any APIRule would be more important, we may also reconsider if we really need the check for APIRule referring to Kyma Gateway at all, since any APIRule check will basically overwrite it in this case, what do you think ?

internal/reconciliations/gateway/dns_entry.go
videlov: This is kind of a mistake I guess.

internal/reconciliations/gateway/dns_entry_test.go
videlov: Some kind of wrong replace I guess. No idea how/why.

internal/reconciliations/gateway/kyma_gateway_test.go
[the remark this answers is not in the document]
videlov: Ahh, I understand the name wrongly then I guess. I was thinking it is checking non-gardener case and not non-gardener resources. Doesn't matter, will bring original names and think of minimising code duplications without the isGardener boolean.

controllers/operator/apigateway_controller.go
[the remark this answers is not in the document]
triffer: Ah I understand, but I think we can still check that the resource blocking is in place for Kyma Gateway, by testing it with Virtual Service blocking the deletion ( this test already exists). And then rely on unit testing for APIRule also blocking the deletion of Kyma Gateway. I think in this case it would be okay to not have an integration test for all possible blocking resources.
Additionally, I think we should keep the logic for blocking Gateway.deletion by APIRule-usage, since this will also be relevant once we allow customer created gateways managed by APIGateway CR. So for Kyma Gateway it might not be so relevant, but for future implementation of gateway management it will be. So the current concrete implementation will be made more generic.

internal/reconciliations/gateway/kyma_gateway.go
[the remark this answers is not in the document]
triffer: I think correctly used it could look like this, to have the key value pairs.
```
ctrl.Log.Info("Custom resource is blocking Kyma Gateway deletion", "gvk", res.GVK.String(), "namespace", res.Namespace, "name", res.Name))
```

controllers/operator/suite_test.go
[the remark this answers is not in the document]
triffer: Ok, I get it. Let's keep it like this for now.

controllers/operator/apigateway_controller.go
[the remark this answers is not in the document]
videlov: Make sense, will move CR finalizer reconciliation as first and adapt the tests.

triffer (approved):

nataliasitko (approved):

Issue #532, opened 2023-09-15
Support for Kyma default gateway in API Gateway module


strekm:
**Description**

APIGateway CR has root property called `enableKymaGateway`, which is by default set to true. There are 2 scenarios that needs to be handled: garden and rest. 
On gardener cluster gardener domain be should read from shoot info and used for setting up Istio Gateway CR. Gardener Certificate CR should be created and generated secret should be referenced in Gateway CR. DNSEntry CR should be created with IP / host from Istio ingressgateway SVC. 
On other clusters static certificate should be used to create a secret referenced in Gateway CR. Static local domain should be used for setting up Gateway CR.

Option to provide configuration by user is not supported anymore. In the future user will be able to provide own default domain and create own gateways (out of the scope of this task).

When `enableKymaGateway` is set to false or completely removed, all related resources mentioned above should be deleted. This should be guarded and if any APIRule is present APIGateway CR should be set to error. In the future this logic should be improved.

ACs:
- [x] default gateway set up properly on gardener and other clusters
  - [x] Add Gardener DNSEntry
  - [x] Add Gardener Certificate
  - [x] Add static Certificate in Secret for non-Gardener
  - [x] Suppot for IP-based and DNS-based load balancer services
- [x] gateway removed is there are no APIRules present and `enableKymaGateway` is set to false or removed
- [x] Kyma gateway removal is blocked by existing APIRules and the status is updated to error (discuss, since warning might be better)
- [x] integration test for k3d and gardener
- [x] documentation 

PR: 
- https://github.com/kyma-project/api-gateway/pull/537
- https://github.com/kyma-project/api-gateway/pull/582

**Reasons**

modularisation

**DoD:**
- [x] provide documentation
~~- [ ] release notes and What's New updates for Kyma customers~~
- [x] provide unit tests
- [x] provide integration tests
- [x] test on production-like environment
~~- [ ] verify resource limits~~
~~- [ ] followup issue~~
~~- [ ] create release and bump in kyma~~
- [x] PR reviewer will verify code coverage and evaluate if it is acceptable

**Attachments**
part of: https://github.com/kyma-project/api-gateway/issues/130
https://github.com/kyma-project/kyma/tree/main/resources/certificates
https://github.com/kyma-project/kyma/blob/main/resources/istio-resources/templates/gateway.yaml

triffer:
There are some open questions:
[...]

PR #582, opened 2023-10-03
Default gateway integration tests

werdes72:
**Description**

- Add integration tests for default kyma gateway

**Related issue(s)**
#532


tests/integration/pkg/resource/manager.go
```diff
@@ -17,6 +17,9 @@ import (
 	"k8s.io/client-go/restmapper"
 )
 
+// TestGatewayOperatorName contains name for APIGateway operator resource
```
triffer, line 20: Having the `TestGatewayOperatorName` in this package feels wrong. What's the reason why it's here?

tests/integration/pkg/resource/manager.go
```diff
@@ -150,6 +233,24 @@ func (m *Manager) UpdateResource(client dynamic.Interface, resourceSchema schema
 	}, m.retryOptions...)
 }
 
+// UpdateResourceWithoutNS updates a given k8s resource without namespace
+func (m *Manager) UpdateResourceWithoutNS(client dynamic.Interface, resourceSchema schema.GroupVersionResource, name string, updateTo unstructured.Unstructured) error {
+	return retry.Do(func() error {
+		time.Sleep(5 * time.Second) //TODO: delete after waiting for resource creation is implemented
```
triffer, line 239: There is still a TODO. What do we do with this?

tests/integration/testsuites/gateway/features/gateway.feature
triffer, line 1: This is about Kyma Gateway, maybe it's better to call the file `kyma_gateway.feature`.

tests/integration/testsuites/gateway/features/gateway.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Checking default kyma gateway
+  Scenario: Kyma gateway is deployed
```
triffer, line 2: This scenario is specific for open source Kyma, right? I mean the resources also apply for Gardener, but in my opinion tests are also documentation of the behaviour. With this test it's not clearly visible, that only k3d must have only this resources.

tests/integration/testsuites/gateway/features/gateway.feature
```diff
@@ -0,0 +1,12 @@
...
+    Given there is an APIGateway operator in "Ready" state
```
triffer, line 3: We don't apply the APIGateway, nor do we remove it after a test. This means we don't know how our APIGateway is configured. The precondition is not part of the Setup of the scenario. I'd not have a central setup APIGateway for all tests. That can easily cause tests being dependant on some order or being flaky. We can also not be sure that the test runs correctly, since we don't start from a "fresh" APIGateway for this scenario.

tests/integration/testsuites/gateway/features/gateway.feature
```diff
@@ -0,0 +1,12 @@
...
+    Then there is a "kyma-gateway" gateway in "kyma-system" namespace
+    And there is a "kyma-gateway-certs" secret in "istio-system" namespace
+
+  Scenario: Kyma gateway cannot be disabled when there is an APIRule
```
triffer, line 7: The scenario describes a simple scenario, when the actual test does way more.
Also the test doesn't fully verify, that kyma-gateway is not removed. It verifies that the CR is in warning state:
`Then disabling kyma gateway will result in "Warning" due to existing APIRule`, but the kyma-gateway could still be removed from the cluster.
I'd also separate this tests to have a clear scenario that is tested.

tests/integration/testsuites/gateway/features/gateway_gardener.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Checking default kyma gateway on gardener
+  Scenario: Kyma gateway is deployed
+    Given there is an APIGateway operator in "Ready" state
```
triffer, line 3: It's the CR, not the operator.
```suggestion
    Given there is an APIGateway CR in "Ready" state
```

tests/integration/testsuites/gateway/features/gateway.feature
```diff
@@ -0,0 +1,12 @@
...
+    Given there is an "kyma-rule" APIRule
+    Then disabling kyma gateway will result in "Warning" due to existing APIRule
+    And enabling kyma gateway back will result in "Ready" state
+    And removing an APIRule and APIGateway operator will also remove "kyma-gateway" in "kyma-system" namespace
```
triffer, line 11: It's the CR, not the operator.
```suggestion
    And removing an APIRule and APIGateway CR will also remove "kyma-gateway" in "kyma-system" namespace
```

tests/integration/testsuites/gateway/features/gateway.feature
triffer, line 3:
```suggestion
    Given there is an APIGateway CR in "Ready" state
```

tests/integration/testsuites/gateway/scenario.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func initScenario(ctx *godog.ScenarioContext, ts *testsuite) {
+	scenario, err := createScenario(ts)
+
+	if err != nil {
+		log.Fatalf("could not initialize custom domain endpoint err=%s", err)
+	}
+
+	ctx.Step(`^there is an APIGateway operator in "([^"]*)" state$`, scenario.thereIsAnAPIGatewayOperator)
```
triffer, line 38: It's the CR, not the operator. We should also adopt the function name in this scenario
```suggestion
	ctx.Step(`^there is an APIGateway CR in "([^"]*)" state$`, scenario.thereIsAnAPIGatewayOperator)
```

tests/integration/testsuites/gateway/scenario.go
```diff
@@ -0,0 +1,279 @@
...
+	ctx.Step(`^there is a "([^"]*)" gateway in "([^"]*)" namespace$`, scenario.thereIsAGateway)
+	ctx.Step(`^there is a "([^"]*)" secret in "([^"]*)" namespace$`, scenario.thereIsACertificate)
+	ctx.Step(`^there is an "([^"]*)" APIRule$`, scenario.thereIsAnAPIRule)
+	ctx.Step(`^disabling kyma gateway will result in "([^"]*)" due to existing APIRule$`, scenario.gatewayErrorWhenKymaGatewayDisabled)
+	ctx.Step(`^enabling kyma gateway back will result in "([^"]*)" state$`, scenario.gatewayReadyWhenEnabled)
+	ctx.Step(`^removing an APIRule and APIGateway operator will also remove "([^"]*)" in "([^"]*)" namespace$`, scenario.removingAPIRuleUnblocksGatewayDeletion)
```
triffer, line 44: It's the CR, not the operator.
```suggestion
	ctx.Step(`^removing an APIRule and APIGateway CR will also remove "([^"]*)" in "([^"]*)" namespace$`, scenario.removingAPIRuleUnblocksGatewayDeletion)
```

tests/integration/testsuites/gateway/features/gateway_gardener.feature
triffer, line 1: This is about Kyma Gateway, maybe it's better to call the file kyma_gateway_gardener.feature.

tests/integration/testsuites/gateway/features/gateway.feature
werdes72: This scenario if for both OS and Gardener.

tests/integration/pkg/resource/manager.go
werdes72: It's also in other functions so I guess waiting for resources is not implemented yet

tests/integration/pkg/resource/manager.go
werdes72: It's used with manager functions, and should be accessible everywhere (both in test scenarios as well as in `main_test.go`). Any suggestion where to move it?

triffer: I think it should be moved to main_test, since this is the only correct usage. It's used in the gateway testsuite for now, but once this is correctly implemented and will apply the APIGateway, the name of the gateway should be part of the creation step and therefore this doesn't need to be referenced. 
Also the code in main_test.go should be moved to a separate file in the future, since it'S only relevant for Before and After testsuite code. But this will be done in a later implementation as far as I understood..

tests/integration/testsuites/gateway/features/gateway.feature
triffer: I don't agree, it's only by accident also working for Gardener, because some of the resource are the same and Open Source doesn't have any additional resources. From a test perspective we are not clearly separating Gardener and Open source test cases. I don't like it.

tests/integration/pkg/resource/manager.go
triffer: Also the name of the variable is not correct. It's not the operator, it is the custom resource.

werdes72: Moved to `config.go`

tests/integration/testsuites/gateway/features/gateway.feature
werdes72: Added separate feature for k3d

triffer (approved):

PR #621, opened 2023-10-13
fix bug in workflow

kolodziejczak:
**Description**

- Remove non existing reference in the workflow


strekm (approved):
/meow

strekm (approved):

PR #625, opened 2023-10-13
Make small corrections to the tutorials

nataliasitko:
**Description**

- Fix rendering of the tabs in OAuth2 tutorial
- Make small language improvements

### DIFF
--- a/docs/user/tutorials/01-50-expose-and-secure-a-workload/01-50-expose-and-secure-workload-oauth2.md
+++ b/docs/user/tutorials/01-50-expose-and-secure-a-workload/01-50-expose-and-secure-workload-oauth2.md
@@ -4,8 +4,8 @@ This tutorial shows how to expose and secure Services using APIGateway Controlle
 
 ## Prerequisites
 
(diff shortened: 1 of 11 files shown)


docs/user/tutorials/01-50-expose-and-secure-a-workload/01-50-expose-and-secure-workload-oauth2.md
```diff
@@ -49,43 +49,52 @@ This tutorial shows how to expose and secure Services using APIGateway Controlle
...
-      </details>
-   </div>
 
+    <div tabs name="export-values">
+    <details>
+    <summary>
+    Token with `read` scope
+    </summary>
+      
+    * Export the following value as an environment variable:
```
NHingerl, line 59: I'm assuming the order of these steps is important, so I'd suggest an ordered list.

```suggestion
    1. Export the following value as an environment variable:
```

If you do that, all indented lines probably need an additional blank space, because numbered lists need 3 instead of 2...

PR #647, opened 2023-10-20
disable deleting and creating APIGateway CR

kolodziejczak:
**Description**

- Disable creation and removal of APIGateway CR

### DIFF
--- a/config/ui-extensions/apigateways/general
+++ b/config/ui-extensions/apigateways/general
@@ -6,5 +6,9 @@ urlPath: apigateways
 category: Kyma
 name: APIGateway
 scope: cluster
+features:
+  actions:
+    disableCreate: true
+    disableDelete: true
 description: >-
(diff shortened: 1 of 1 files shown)


config/ui-extensions/apigateways/general
```diff
@@ -6,5 +6,9 @@ urlPath: apigateways
 category: Kyma
 name: APIGateway
 scope: cluster
+features:
+  actions:
+    disableCreate: true
```
barchw, line 11: Just for me to understand, does that mean that user needs to create the apigateway CR from e.g. terminal for the first time?

kolodziejczak: it's created by Lifecycle Manager

barchw: I guess OS is not a problem, as the installation documentation mentions creating CR from CLI anyway

barchw (approved):

Issue #465, opened 2023-08-22
Object already has been modified error


kolodziejczak:
**Description**

During executing integration tests errors are thrown about updating object that already has been modified in K8S.

**Expected result**

Check how to reproduce it manually, and then apply appropriate fixes to eliminate the error.

**Actual result**

Object has been already modified error thrown during integration tests.

**Steps to reproduce**
Run integration tests on local k3d cluster.

PR:
https://github.com/kyma-project/api-gateway/pull/695

PR #695, opened 2023-11-03
Fix object already modified error

werdes72:
**Description**

- Get an ApiRule from the cluster just before updating it's status, preventing object already modified errors

**Related issue(s)**
#465

### DIFF
--- a/controllers/gateway/apirule_controller.go
+++ b/controllers/gateway/apirule_controller.go
@@ -286,6 +286,11 @@ func retryReconcile(err error) (ctrl.Result, error) {
 }
 
 func (r *APIRuleReconciler) updateStatus(ctx context.Context, api *gatewayv1beta1.APIRule, status processing.ReconciliationStatus) (*gatewayv1beta1.APIRule, error) {
+	api, err := r.getLatestApiRule(ctx, api)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


controllers/gateway/apirule_controller.go
```diff
@@ -295,9 +300,25 @@ func (r *APIRuleReconciler) updateStatus(ctx context.Context, api *gatewayv1beta
 	api.Status.AuthorizationPolicyStatus = status.AuthorizationPolicyStatus
 
 	r.Log.Info("Updating ApiRule status", "status", api.Status)
-	err := r.Client.Status().Update(ctx, api)
+	err = r.Client.Status().Update(ctx, api)
```
videlov, line 303: can we have a retry in here, like in Istio:
https://github.com/kyma-project/istio/blob/88aa69cd79b7ca3cf788e8a038ca8bc851809eef/controllers/status.go#L28-L45

we could retry setting the status, not needed to requeue whole reconcile for it when it fails:
https://github.com/kyma-project/api-gateway/blob/131495f2086a9b7a0ff896cb5f5768824ea1989f/controllers/gateway/apirule_controller.go#L246

videlov (changes requested):

barchw:
I checked as well, and the error does no longer happen :+1

videlov (approved):

PR #751, opened 2023-11-28
Oathkeeper Maester resources first-time reconciliation failure

triffer:
**Description**

- Change order of maester resources 
- Change criteria to consider Oathkeeper deployment to be ready. We cannot use `dep.Status.UnavailableReplicas > 0`, since sometimes this passes, even when the deployment isn't ready. This happens probably, because the default value of UnavailableReplicas in the status struct is `0`.
- Use the `Available` condition type to verify the readiness of the Oathkeeper deployment. This also requires some tests to be adapted, since the status is not updated in these tests.
- Reduce retry for CRD and maester cluster role.

**Related issue(s)**


internal/reconciliations/oathkeeper/deployment.go
```diff
@@ -133,3 +128,13 @@ func getReplicasForDeployment(ctx context.Context, k8sClient client.Client) (str
 
 	return strconv.Itoa(int(*dep.Spec.Replicas)), nil
 }
+
+func hasMinimumReplicasAvailable(dep appsv1.Deployment) bool {
```
barchw, line 132: With the new approach we want wait for the restart to finish. As so, for example a CrashLoopBackoff of new pods during a rollout restart will not be seen as an error (as the deployment is in ConditionTrue during the whole restart/upgrade). Only the ReplicaSet condition is in progressing state. Are we fine with that?

barchw (changes requested):

internal/reconciliations/oathkeeper/deployment.go
triffer: I think with waiting for the deployment to be available we are already crossing the line of responsibility of the operator a little bit. 
In my opinion the operator has the responsibility to reconcile the cluster to an expected state. In our case one part of it is, that Oathkeeper resources are applied. Additionally we wait for the deployment to be Ready/Available. This validation of availability can already be seen as part of the monitoring. And monitoring is not a good use case for the reconciliation, because it succeeded, the next default reconciliation will happen in approx. 10h and until than no "monitoring" will happen.
If we not only monitor the availability of the deployment, but also the status of the ReplicaSet, we go even deeper into misusing reconciliation for monitoring.

barchw: Yes, in that terms it makes sense.

PR #780, opened 2023-12-20
Fix watch issue with Ory Rules

barchw:
**Description**

Listing an object using structured interface causes k8s client-go to register a `reflector` with https://github.com/kubernetes/client-go/blob/fb8b7346aacefea5ee2ab2e234afc4451c90c435/tools/cache/reflector.go#L208

This behaviour is not intended in our scenario, as we are expecting that we might be removing the rule CRD.

- Use unstructured to list Ory Rules

**Related issue(s)**


controllers/operator/apigateway_controller.go
```diff
@@ -246,7 +245,13 @@ func apiRulesExist(ctx context.Context, k8sClient client.Client) (bool, error) {
 }
 
 func oryRulesExist(ctx context.Context, k8sClient client.Client) (bool, error) {
-	oryRulesList := oryv1alpha1.RuleList{}
+	oryRulesList := unstructured.UnstructuredList{}
```
triffer, line 248: Can we add a comment why we use the UnstructuredList?
```suggestion
         // To prevent reconciliation errors during Oathkeeper uninstallation, which arise from the absence of the Oathkeeper CRD,
	// we exclude calls to the rules from utilizing the cache. This can be achieved by using unstructured objects, as they are excluded from caching.
	oryRulesList := unstructured.UnstructuredList{}
```

triffer (approved):

PR #857, opened 2024-02-06
Integration tests with OAuth2 mock

triffer:
**Description**

- Add OAuth2 mock in integration tests

**Related issues**


tests/integration/pkg/hooks/api-gateway.go
```diff
@@ -22,7 +23,7 @@ import (
 )
 
 const templateFileName string = "pkg/hooks/manifests/apigateway_cr_template.yaml"
-const ApiGatewayCRName string = "test-gateway"
+const ApiGatewayCRName string = "default"
```
triffer, line 26: I would keep the separate name for the integration test gateway, maybe even add a scope by naming it `integration-test`. 
When this has a different name then it will fail early when you run it on a cluster where api-gateway is already installed. Otherwise it might lead to unexpected behaviour, if you are not using a clean api-gateway installation for the integration tests.

tests/integration/pkg/hooks/manifests/apigateway_cr_template.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: operator.kyma-project.io/v1alpha1
 kind: APIGateway
 metadata:
-  name: test-gateway
+  name: default
```
triffer, line 4: We have this duplicated in with the code in `api-gateway.go`. IMP it's enough to have it hardcoded in this file. We also have this `apigateway_cr_template.yaml` and a `api-gateway.yaml` with actual name templating next to it, that seems to be unused. We should clean that up.

tests/integration/pkg/hooks/api-gateway.go
barchw: I thought that as we are discouraging names other than `default`, it is not a good solution to test our integration with something different. 
And to be honest, I had a counter effect to this, meaning that my tests were failing locally since I had a `default` API Gateway CR created, and was wondering why does it fail

triffer: Ok, good points, then let's keep default.

triffer:
LGTM

barchw (approved):

nataliasitko (approved):

Issue #832, opened 2024-01-24
Support mTLS in default certificate 


strekm:
**Description**

API Gateway module comes with default domain, simple TLS gateway and certificate. To support mTLS in early development phase add *.mtls.{kyma-domain} to DNS Names of issued default Certificate CR. This way developers can reuse this certificate to implement in their own mTLS gateway. This would ease setup of mTLS in Kyma. Managed Kyma users can reuse Kyma default domain and certificate in development. This approach should not be used in production setup. Following scenario should be possible:
- enhance default Kyma certificate with additional root certificate 
- create mTLS gateway for *.mtls.{kyma-domain} reusing default Kyma certificate
- expose httpbin on httpbin.mtls.{kyma-domain}
- call httpbin workload with valid certificate
- inspect XFCC-header on httpbin side

TODO:
- [ ] dbl check if dns names is needed (if mtls dns name is need implement, otherwise nothing to do here)
- [ ] check staging and production envs for usage of *.mtls.{kyma-domain}
- [ ] consult @TorstenD-SAP on reusing Kyma default certificate

ACs:
- [ ] developer can setup mTLS gateway reusing default Kyma certificate
- [ ] developer can add secret with root certificate
- [ ] simple TLS gateway is not changed
- [ ] note in documentation that this is for development purposes only

**Reasons**

Easy implementation of mTLS in Kyma.

**DoD:**
- [x] Provide unit and integration tests.
- [x] Provide documentation.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed.
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [x] Add release notes.

**Attachments**

Ressetkk:
Okay, so I'm going to sum up my R&D work around supporting mTLS.
I've set up some additional configuration and confirmed, that kyma-tls certificate can be indeed extended. However I've noticed some additional configuration that is applied around default gateway that api-gateway creates. In managed offering additionally there is already self-signed cert created outside of the bounds of api-gateway default gateway. I've confirmed it with @ebensom and @gczuczy.

To make mTLS possible in basic spec some additional configuration is needed:
* Create (or buy) Root CA certificate
* Create and sign server certificate and add it as secret
* Create and sign client certificates
* Create gateway, or extend existing gateway with a server pointing to some domain, with server certificate
[...]

PR #867, opened 2024-02-12
Add mTLS guide for istio and api-gateway

Ressetkk:
/area ci
/kind feature

Updated documentation for mTLS-enabled gateway with step-by-step guide on getting required certificates for client cert validation

#832


docs/user/tutorials/01-30-set-up-mtls-gateway.md
```diff
@@ -0,0 +1,134 @@
...
+
+The mTLS Gateway is exposed under `*.mtls.example.com` with valid DNS `A` record.
+
+## Create DNS Entry and generate wildcard certificate
+
+> Note: This step is heavily dependent on the configuration of a hyperscaler. Always consult official documentation of each cloud service.
+
+For Gardener shoot clusters follow ["set up a custom domain for workload" guide](01-10-setup-custom-domain-for-workload.md).
+
+## Generate self-signed Root CA and client certificate
```
nataliasitko, line 25: We also have the tutorial [Prepare Self-Signed Root Certificate Authority and Client Certificates](https://kyma-project.io/#/api-gateway/user/tutorials/01-60-security/01-61-mtls-selfsign-client-certicate?id=prepare-self-signed-root-certificate-authority-and-client-certificates). If it's up to date, we could link it in prerequisites instead of adding this step

Ressetkk: Looks like it's the same thing and could be reused

docs/user/tutorials/01-30-set-up-mtls-gateway.md
```diff
@@ -0,0 +1,124 @@
...
+EOF
+```
+
+## Create secret containing Root CA cert
+
+In order for `MUTUAL` mode working correctly, a Root CA must be applied on a cluster and follow [Istio naming convention](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings) to be used by it.
+Create an Opaque secret containing previously generated Root CA certificate:
+
+```sh
+    kubectl create secret generic -n default kyma-mtls-certs-cacert --from-file=cacert=cacert.crt
```
triffer, line 65: If I remember correctly, secrets will be only read from the namespace where the ingress gateway is deployed. So this should probably also be created in `istio-system`.

docs/user/tutorials/01-30-set-up-mtls-gateway.md
```diff
@@ -0,0 +1,124 @@
...
+cat <<EOF | kubectl apply -f -
+apiVersion: gateway.kyma-project.io/v1beta1
+kind: APIRule
+metadata:
+  labels:
+    app.kubernetes.io/name: httpbin-mtls
+  name: httpbin-mtls
+  namespace: default
+spec:
+  gateway: kyma-mtls-gateway.default.svc.cluster.local
```
triffer, line 84: I think we use the form of `<gateway namespace>/<gateway name>` in other tutorials and this is also what is mejn tioned in Istio documentation, so it might be a good idea to use the same format in this tutorial.

docs/user/tutorials/01-30-set-up-mtls-gateway.md
Ressetkk: this secret needs to be created under the same namespace where the certificate secret has been provisioned. Your certificate will not always be in the Istio-system.

triffer: As far as I remember, your secret needs to be in istio-system namespace if you want ingress gateway to use this secret. This is a limitation that was only solved recently for new Kubernetes Gateway API.

Ressetkk: Good point. I've updated the docs.

triffer (approved):

nataliasitko (approved):

Issue #866, opened 2024-02-12
Validation of noop access strategy does not match the documentation


triffer:
**Description**

In our [APIRule docs](https://github.com/kyma-project/api-gateway/blob/main/docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md) we mention:

> Access strategies noop or allow cannot be used with any other access strategy on the same spec.rules.path.

While this is true for allow, that's not the case for noop .

We don't have any validation in place and it's possible to configure noop and jwt handler for the same path. In this case access is always possible.

**Expected result**
When noop and another access strategy is on the same path, the validation will fail with message
```
noop access strategy is not allowed in combination with other access strategies
```

**Actual result**
`noop` handler can be added with other handlers on the same path and changes the behaviour, e.g. by allowing access to JWT secured path.

**Steps to reproduce**
- Deploy a workload
- Secure workload with APIRule that exposes `GET`and `POST` on path `/anything` with jwt and noop handler.
```
kubectl apply -f - <<EOF
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: httpbin-rule
  namespace: default
  labels:
    app: httpbin
spec:
  gateway: kyma-system/kyma-gateway
  host: httpbin
  service:
    name: httpbin
    port: 8000
  rules:
    - path: /anything
      methods: [ "GET", "POST"]
      mutators: [ ]
      accessStrategies:
        - handler: jwt
          config:
            trusted_issuers:
              - https://kymagoattest.accounts400.ondemand.com
            jwks_urls:
              - https://kymagoattest.accounts400.ondemand.com/oauth2/certs
        - handler: noop
EOF
```
- APIRule in status OK
- Access to GET on `/anything` is not restricted

**Troubleshooting**

Issue #329, opened 2023-04-26
[ADR] Supported API-Gateway configurations


kasiakepka:
# Supported API-Gateway configurations

Date: 2022-12-08

## Status

Accepted

## Context

While having different configuration options for APIRule we specify which combinations of `path` + `accessStrategies.handler` are allowed for API-Gateway component. Additionally we will provide validation based on this.

More information on the current test coverage and validation you may find on:
https://github.tools.sap/xf-goat/knowledge-hub/tree/master/api-gateway-handlers

## Decision

After having a meeting today and discussed the issue within the team, we decided on the following:

### Handlers matrix based on a rule with 1 path and any 2+ handlers

| First/Second handler | allow | noop | oauth2 | jwt (ory) | jwt (istio) |
|:---:|:---:|:---:|:---:|:---:|:---:|
| **allow** | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) |
| **noop** | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) |
| **oauth2** | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) |
| **jwt (ory)** | ![](https://placehold.co/15x15/f03c15/f03c15.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/00FF00/00FF00.png) | ![](https://placehold.co/15x15/f03c15/f03c15.png) |
[...]

PR #915, opened 2024-03-01
Validate secure and unsecure access strategies

werdes72:
**Description**

- Validate usage of secure and unsecure access strategies at the same time
- Integration tests

**Related issues**
#866


internal/validation/validate.go
```diff
@@ -255,6 +256,17 @@ func (v *APIRuleValidator) validateAccessStrategies(attributePath string, access
 	for i, r := range accessStrategies {
 		strategyAttrPath := attributePath + fmt.Sprintf("[%d]", i)
 		problems = append(problems, v.validateAccessStrategy(strategyAttrPath, r, selector, namespace)...)
+		switch r.Handler.Name {
```
triffer, line 259: I think the logic we have implemented here is not correct, because it allows combination that should be restricted, e.g. "allow" and "noop" on same rule. 

I remembered that we have a decision about that. Obviously no one remember about that, when I asked it in Slack and when we were discussing this in the refinement. 😁 I checked the decision we have in [api-gateway](https://github.com/kyma-project/api-gateway/issues/329). And this decision is actual different then what we have in the docs. As far as I understand the decision, noop is allowed to be on the same path with OAuth2 and Ory JWT, but I think this doesn't make sense and is incorrect in the ADR.

Also the validation is different depending on if you are using JWT or ORY handler, so we shouldn't mix this with general validation.
There is already a validation implemented for the other validations in [istio as_validator](https://github.com/kyma-project/api-gateway/blob/main/internal/processing/istio/as_validator.go) and [ory as_validator](https://github.com/kyma-project/api-gateway/blob/main/internal/processing/ory/as_validator.go).

tests/integration/testsuites/istio-jwt/features/istio_jwt.feature
```diff
@@ -44,6 +44,22 @@ Feature: Exposing endpoints with Istio JWT authorization strategy
     And JwtAndUnrestricted: Calling the "/json" endpoint without token should result in status between 200 and 299
     And JwtAndUnrestricted: Teardown httpbin service
 
+  Scenario: Exposing a httpbin endpoint with JWT and no_auth in the same rule
```
triffer, line 47: Maybe we can make this more expressive, so it's easier to understand what general behaviour is tested.
```suggestion
  Scenario: The APIRule is in status Error when there are handlers in the same rule that must not be combined with others
```

tests/integration/testsuites/istio-jwt/features/istio_jwt.feature
```diff
@@ -44,6 +44,22 @@ Feature: Exposing endpoints with Istio JWT authorization strategy
...
+    Given JwtAndNoAuth: There is a httpbin service
+    And JwtAndNoAuth: There is an endpoint secured with JWT and no_auth on path "/ip"
+    And JwtAndNoAuth: There is an endpoint with handler "no_auth" on path "/ip"
+    When JwtAndNoAuth: Create APIRule
+    Then JwtAndNoAuth: The APIRule has "ERROR" status
+    And JwtAndNoAuth: Teardown httpbin service
+
+  Scenario: Exposing a httpbin endpoint with JWT and no_auth on the same path with different methods
```
triffer, line 55: Maybe we can make this more expressive, similar to the other test
```suggestion
  Scenario: The APIRule is in OK status when handlers that may not be combined with others are defined in different rules
```

tests/integration/testsuites/istio-jwt/features/istio_jwt.feature
triffer, line 1: I think we shouldn't do validation tests in the integration tests if it's not really necessary, since they are expensive.
We usually have validation tests in our envtest-based integration tests: https://github.com/kyma-project/api-gateway/blob/main/controllers/gateway/apirule_controller_validation_integration_test.go

triffer (changes requested):

tests/integration/testsuites/istio-jwt/features/istio_jwt.feature
werdes72: Moved the tests

internal/validation/access_strategy.go
```diff
@@ -25,3 +25,24 @@ func CheckForExclusiveAccessStrategy(accessStrategies []*gatewayv1beta1.Authenti
 
 	return nil
 }
+
+// CheckForSecureAndUnsecureAccessStrategies checks if there are secure and unsecure access strategies used at the same time.
+func CheckForSecureAndUnsecureAccessStrategies(accessStrategies []*gatewayv1beta1.Authenticator, attributePath string) []Failure {
```
triffer, line 30: Is there a reason why we didn't add unit tests on package level? Might be confusing for someone that changes the functionality of a method like this.

triffer:
I think we also need to update the documentation to mention the change of allowed access strategies. I think it would also be good to have release notes for the bugfix of blocking secured and unsecured handlers in the same rule.

internal/validation/access_strategy.go
werdes72: Added unit tests.

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
```diff
@@ -42,6 +42,7 @@ This table lists all parameters of APIRule CRD together with their descriptions:
 > [!WARNING]
 >  Having both the Oathkeeper and Istio `jwt` access strategies defined is not supported. Access
 > strategies `noop`, `allow`, or `no_auth` cannot be used with any other access strategy on the same **spec.rules.path**.
+> Mixing secure and unsecure access strategies is not supported.
```
nataliasitko, line 45:
```suggestion
> You must not use the access strategies `jwt`, `oath2_introspection`, `noop`, `allow`, or `no_auth` with any other access strategy on the same **spec.rules.path**.
> Having both the Oathkeeper and Istio `jwt` access strategies defined is not supported.
```
By 'secure access strategies' we mean `jwt` and `oath2_introspection`, right? I assume that there still are cases when the user can define more than two access strategies

nataliasitko, line 45:
```suggestion
On the same **spec.rules.path**, you must not:
  - use the access strategies `jwt`, `oath2_introspection`, `noop`, `allow`, or `no_auth` with any other access strategy
  - define both the Oathkeeper and Istio `jwt` access strategies
  - combine secure and unsecure access strategies. The secure access strategies include `jwt` and `oath2_introspection`.
```
By 'secure access strategies' we mean `jwt` and `oath2_introspection`, right? Are there any cases when we do support defining more than one access strategy?

docs/release-notes/2.3.0.md
```diff
@@ -1,3 +1,7 @@
 
+## New features
+- Add a validation that forbids setting up secure and unsecure handlers on the same path [#915](https://github.com/kyma-project/api-gateway/pull/915)
```
triffer, line 3: Should we "sell it" as a feature? I think it's more like a bugfix for bad validation. But maybe that's just me ... 😁

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
triffer: > use the access strategies `jwt`, `oath2_introspection`, `noop`, `allow`, or `no_auth` with any other access strategy.

This is not correct, you can use jwt and oauth2_introspection in the same rule.
For the Ory handler "noop", "no_auth" and "allow" can't be used with any other access strategy.
For istio handler "jwt", "noop", "no_auth" and "allow" can't be used with any other access strategy.

nataliasitko:
```suggestion
> When you use the Ory handler, do not define the access strategies `noop`, `allow`, or `no_auth` with any other access strategy on the same **spec.rules.path**.
> When you use the Istio handler, do not define the access strategies `jwt`, `noop`, `allow`, or `no_auth` with any other access strategy on the same **spec.rules.path**..
```
Thanks, would this be fine?

internal/validation/access_strategy.go
```diff
@@ -25,3 +25,24 @@ func CheckForExclusiveAccessStrategy(accessStrategies []*gatewayv1beta1.Authenti
...
+	var containsSecureAccessStrategy, containsUnsecureAccessStrategy bool
+
+	for _, r := range accessStrategies {
+		switch r.Handler.Name {
+		case gatewayv1beta1.AccessStrategyOauth2ClientCredentials, gatewayv1beta1.AccessStrategyOauth2Introspection, gatewayv1beta1.AccessStrategyJwt:
+			containsSecureAccessStrategy = true
+		case gatewayv1beta1.AccessStrategyNoop, gatewayv1beta1.AccessStrategyNoAuth, gatewayv1beta1.AccessStrategyAllow,
+			gatewayv1beta1.AccessStrategyUnauthorized, gatewayv1beta1.AccessStrategyAnonymous, gatewayv1beta1.AccessStrategyCookieSession:
```
triffer, line 38: I just saw that, why is CookieSession considered unsecure?

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
nataliasitko: I see. I edited the second suggestion and included the new validation

internal/validation/access_strategy.go
werdes72: Fixed

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
werdes72: Docs updated

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
```diff
@@ -40,8 +40,9 @@ This table lists all parameters of APIRule CRD together with their descriptions:
 >  If `service` is not defined at the **spec.service** level, all defined Access Rules must have `service` defined at the **spec.rules.service** level. Otherwise, the validation fails.
 
 > [!WARNING]
->  Having both the Oathkeeper and Istio `jwt` access strategies defined is not supported. Access
-> strategies `noop`, `allow`, or `no_auth` cannot be used with any other access strategy on the same **spec.rules.path**.
+> When you use the Ory handler, do not define the access strategies `noop`, `allow`, or `no_auth` with any other access strategy on the same **spec.rules.path**.
+> When you use the Istio handler, do not define the access strategies `jwt`, `noop`, `allow`, or `no_auth` with any other access strategy on the same **spec.rules.path**.
+> Additionally, do not use secured access strategies (such as `jwt`, `oauth2_introspection`, or `oauth2_client_credentials`) with unsecured access strategies (for example, `allow`, `no_auth`, `noop`, `unauthorized`, or `anonymous`).
```
triffer, line 45: I think we should have the complete list.
```suggestion
> Additionally, do not use secured access strategies (such as `jwt`, `oauth2_introspection`, `oauth2_client_credentials`, or `cookie_session`) with unsecured access strategies (for example, `allow`, `no_auth`, `noop`, `unauthorized`, or `anonymous`).
```

triffer (approved):

internal/validation/access_strategy_test.go
```diff
@@ -96,4 +96,110 @@ var _ = Describe("Access Strategies Validation", func() {
...
+					Handler: &gatewayv1beta1.Handler{
+						Name: gatewayv1beta1.AccessStrategyUnauthorized,
+					},
+				},
+				{
+					Handler: &gatewayv1beta1.Handler{
+						Name: gatewayv1beta1.AccessStrategyAnonymous,
+					},
+				},
+				{
```
triffer, line 156: This is no unsecure access strategy.

werdes72: Fixed

triffer (approved):

nataliasitko (approved):

Issue #907, opened 2024-02-29
Conditions implementation


strekm:
**Description**

Enhance API Gateway CR Status specification with conditions. Conditions should be implemented based on proposal and ADR. Changes should be tested automatically. API Gateway CRD documentation should be extended to include condition types and reasons. UI of API Gateway should display conditions in consistent way to Istio CR and other modules. 

ACs:
- [ ] API Gateway CRD extended with conditions
- [ ] API Gateway UI updated
- [ ] documentation updated

**Reasons**
Improve status of API Gateway CR

**DoD:**
- [ ] Provide unit and integration tests.
- [ ] Provide documentation.
- [ ] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [ ] If you changed the resource limits, explain why it was needed.
- [ ] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [ ] Add release notes.

**Attachments**
https://github.com/kyma-project/api-gateway/issues/872

Issue #903, opened 2024-02-26
[ADR] API Gateway Custom Resource Status Improvements


videlov:
# API Gateway Custom Resource Status Improvements

## Status
Accepted

## Context
The API Gateway CR status doesn't contain enough information to easily debug without directly accessing Kyma API Gateway Operator's logs. We would like to make it easier both for the user and the Kyma API Gateway team to view the state of the API Gateway module.

## Decision

For the sake of simplicity, we implement the condition of type `Ready` as the main condition that supports multiple reasons. This way, we can reduce the number of condition types. If we use the [SetStatusCondition function](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition) of [k8s.io/apimachinery](http://k8s.io/apimachinery), the **lastTransitionTime** is only updated when the status of the condition changes.

To support multiple reasons for the status `False`, we decided to set the status of the `Ready` condition to `Unknown` at the beginning of the reconciliation. This allows us to force the **lastTransitionTime** of the `Ready` condition to be always updated.

In addition, the `Ready` condition reflects one reconciliation status. Therefore, it is a good idea to set it to `Unknown` at the beginning.

For `KymaGatewayDeletionBlocked` and `DeletionBlockedExistingResources` reasons of the `Ready` condition with the status `Warning` we report up to 5 custom resources that are blocking the deletion. The listing format for each is `<kind>/<name>`.

Conditions:

| CR state   | Type                         | Status | Reason                                | Message                                                                                        |
|------------|------------------------------|--------|---------------------------------------|------------------------------------------------------------------------------------------------|
| Ready      | Ready                        | True   | ReconcileSucceeded                    | Reconciliation succeeded                                                                       |
| Error      | Ready                        | False  | ReconcileFailed                       | Reconciliation failed                                                                          |
| Error      | Ready                        | False  | OlderCRExists                         | API Gateway CR is not the oldest one and does not represent the module state                   |
| Error      | Ready                        | False  | CustomResourceMisconfigured 			 | API Gateway CR has invalid configuration                                                       |
| Error      | Ready                        | False  | DependenciesMissing                   | Module dependencies missing                                                                   |
| Processing | Ready                        | False  | KymaGatewayReconcileSucceeded         | Kyma Gateway reconciliation succeeded                                                          |
| Error      | Ready                        | False  | KymaGatewayReconcileFailed            | Kyma Gateway reconciliation failed                                                             |
[...]

PR #921, opened 2024-03-05
feat: add conditions to gateway cr

Ressetkk:
/kind feature
/area api-gateway

The implementation has been done through extending `Status` interface and underlying struct with metav1.Condition field. Every time the `UpdateAPIGatewayStatus` function gets called, it updates the condition in status field to provided condition. As a safeguard added functionality that if the consition is missing, then do not change the condition. The problem might be when APIGateway CRD is not applied on the cluster and there is no CR to be updated. However it's very unlikely that it would happen.

Added `conditions` package that contains custom ReasonMessage struct with default variables that can be used as default conditions passed to `UpdateAPIGatewayStatus()`. Fields in a struct are private and cannot be modified outside package.

Extended functions that return Status interface to include nullable condition argument.

Added test for `UpdateAPIGatewayStatus()` to check if expected condition gets set in the CR status.

I'm not a fan of using `AdditionalMessage` but at the time of implementation I didn't have better idea how to handle message. Maybe structured key/val field and listing it would be better? Until then, it can be left as-is.

#907


controllers/operator/apigateway_controller.go
```diff
@@ -203,49 +205,56 @@ func (r *APIGatewayReconciler) reconcileFinalizer(ctx context.Context, apiGatewa
...
 		if err != nil {
-			return controllers.ErrorStatus(err, "Error during listing existing ORY Oathkeeper Rules")
+			return controllers.ErrorStatus(err, "Error during listing existing ORY Oathkeeper Rules", conditions.ReconcileFailed.Condition())
 		}
-		if oryRulesFound {
+		if len(oryRulesFound) > 0 {
 			return controllers.WarningStatus(errors.New("could not delete API-Gateway CR since there are ORY Oathkeeper Rule(s) that block its deletion"),
-				"There are ORY Oathkeeper Rule(s) that block the deletion of API-Gateway CR. Please take a look at kyma-system/api-gateway-controller-manager logs to see more information about the warning")
+				"There are ORY Oathkeeper Rule(s) that block the deletion of API-Gateway CR. Please take a look at kyma-system/api-gateway-controller-manager logs to see more information about the warning",
+				conditions.OathkeeperReconcileFailed.AdditionalMessage(": "+strings.Join(oryRulesFound, ",")).Condition())
```
videlov, line 230: fyi, https://github.com/kyma-project/api-gateway/issues/903

```
with the status Warning we report up to 5 custom resources that are blocking the deletion. The listing format for each is <kind>/<name>.
```

Ressetkk: I think it won't be meaningful to list only up to 5 resources. I don't think the list you get from apiserver is ordered so you may end up with overriding condition. I would leave listing all conditions and if the logs get cluttered too much, we can adjust it. What do you think?

videlov: What will your condition message be, when you have 1000 blocking resources?

internal/conditions/conditions.go
```diff
@@ -0,0 +1,40 @@
...
+	return &metav1.Condition{
+		Type:    "Ready",
+		Reason:  rm.reason,
+		Message: rm.message,
+		Status:  rm.status,
+	}
+}
+
+func (rm *ReasonMessage) AdditionalMessage(message string) *ReasonMessage {
+	rm.message = fmt.Sprintf(rm.message, message)
```
videlov, line 38: https://go.dev/play/p/S07wsWDWPmn

without a test you just assume it will look good?

controllers/operator/apigateway_controller.go
Ressetkk: 1000 resources! it will generate 1000 log lines though... Maybe we can think about how to handle it differently? Keeping it in condition message seems kind of incorrect to me anyway. If we generate log lines then do we really need it in a condition message?

internal/conditions/conditions.go
Ressetkk: I assumed that message would contain string format and that'd just add the message. Now that I think about it this function will do nothing with current default conditions.

videlov: no test = no approve, I am not reviewing this any further

triffer (commented):
I added some tests during review to verify the functionality instead of testing it manually on the cluster. I will just push it. Some tests also fail right now, because the implementation might not be correct, let's discuss and adapt the tests if we want to have the functionality different.

controllers/operator/apigateway_controller.go
triffer: I'd say the Apigateway CR is our main "communication channel" with the user. Having it in the logs only didn't work in the past. We saw that customers were missing this or wondering why it's so complicated. 
Having a shorter list is somehow a compromise in between the two options and we want to see if this is better for the user. Of course we can revisit this, but I think in the first approach we should implement our what we agreed on and check if this is sufficient.

controllers/status.go
```diff
@@ -30,47 +31,54 @@ type Status interface {
...
 
 type status struct {
 	err         error
 	description string
 	state       State
+	condition   *metav1.Condition
 }
 
-func ErrorStatus(err error, description string) Status {
+func ErrorStatus(err error, description string, condition *metav1.Condition) Status {
```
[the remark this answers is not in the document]
triffer: We keep it like it is right now, maybe we want to change it in the future.

internal/conditions/conditions.go
```diff
@@ -0,0 +1,40 @@
...
+func (rm *ReasonMessage) Condition() *metav1.Condition {
+	return &metav1.Condition{
+		Type:    "Ready",
+		Reason:  rm.reason,
+		Message: rm.message,
+		Status:  rm.status,
+	}
+}
+
+func (rm *ReasonMessage) AdditionalMessage(message string) *ReasonMessage {
```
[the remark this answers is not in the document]
triffer: We have a doc comment and this should be good enough.

triffer:
We need to add the conditions to the Apigateway UI. In the same way it is in [Istio UI](https://github.com/triffer/istio/blob/a920fa2a819c6334986d9d912451a6eba0dfaeab/config/ui-extensions/istios/details#L147).

controllers/operator/apigateway_controller_test.go
```diff
@@ -287,35 +302,45 @@ var _ = Describe("API-Gateway Controller", func() {
...
-				Kind:    "rule",
-			})
-
-			Expect(err).ToNot(HaveOccurred())
-
-			c := createFakeClient(apiGatewayCR, r)
+			}
+			var rules []client.Object
+			for i := 0; i < 6; i++ {
+				// we initialize more than 5 objects, so we validate if we show only 5 in a condition
```
triffer, line 326: I think we are missing the validation, that we show only 5. As far as I can see we only validate that the message contains 5.

controllers/operator/apigateway_controller_test.go
```diff
@@ -254,20 +268,21 @@ var _ = Describe("API-Gateway Controller", func() {
...
-				},
-			}
-			oryRule := &oryv1alpha1.Rule{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "ory-rule",
-					Namespace: "default",
-				},
+
+			var rules []client.Object
+			// we initialize more than 5 objects, so we validate if we show only 5 in a condition
```
triffer, line 273: I think we are missing the validation, that we show only 5. As far as I can see we only validate that the message contains 5.

internal/reconciliations/gateway/gateway.go
```diff
@@ -54,39 +56,46 @@ func ReconcileKymaGateway(ctx context.Context, k8sClient client.Client, apiGatew
...
 
 		if len(clientResources) > 0 {
-			for _, res := range clientResources {
+			var blockingResources []string
+			for i := 0; i < 5; i++ {
+				if i >= len(clientResources) {
+					break
+				}
+				res := clientResources[i]
 				ctrl.Log.Info("Custom resource is blocking Kyma Gateway deletion", "gvk", res.GVK.String(), "namespace", res.Namespace, "name", res.Name)
```
triffer, line 86: Do we need to move this log outside of the loop to be reliable? With the current implementation we will only log the first 5 blocking resources, but we want to have all in the log.

config/ui-extensions/apigateways/details
triffer, line 1: I think it would be good to also remove the tabs to have the same look and feel as in the Istio CR UI.

controllers/operator/apigateway_controller_test.go
Ressetkk: I don't quite get it. If the message contains less than 5, the loop will iterate over all objects. Since the condition message is set based on the size of the slice returned in API, checking number of elements as substring seems completely fine for me. Especially since the idea was to show only up to 5 values if the response has more than 5 elements.

internal/reconciliations/gateway/gateway.go
Ressetkk: Do you? The implementation was only about first 5. If there are going to be 1000 blocking elements, your logs will be pretty much cluttered. I can modify it, so it goes over all elements but only collects 5 elements but I would avoid cluttering the logs.

config/ui-extensions/apigateways/details
Ressetkk: It's been copy-pasted from the Istio UI :D The intendation is exactly the same

triffer (approved):

nataliasitko (approved):

PR #923, opened 2024-03-06
External authorizer api proposal

barchw:
**Description**

- Create external authorizer API


docs/user/technical-reference/external-authorizer.md
```diff
@@ -0,0 +1,160 @@
...
+  jwtRules:
+  - issuer: https://example.com
+    jwksUri: https://example.com/.well-known/jwks.json
+```
+
+As so we should allow combination of this handler with others.
+
+### Support for multiple external authorizers combinations
+
+We need to consider whether a configuration that will use multiple external authorizers on one path is valuable. Technically, this is possible to do, as all CUSTOM policies will need to result in `allow` response for the request to be allowed.
```
triffer, line 65: I would consider this as an absolute corner case and in my opinion we should keep the configuration simple as a first version and only allow one external authorizer per rule.

docs/user/technical-reference/external-authorizer.md
```diff
@@ -0,0 +1,160 @@
...
+
+## API Proposal
+
+Considering the logic of CUSTOM Istio AuthorizationPolicy, suggestion would be to not handle external authorizer as an additional access strategy, but have a separate field (array) in `rule`, that would configure only external authorizer.
```
triffer, line 69: In my opinion it's confusing to detach this from the other access strategies, because from a context they do the same.

triffer:
Maybe we can also add information about what we want to validate for the API, e.g. existence of external authorizer.

docs/user/technical-reference/external-authorizer.md
```diff
@@ -0,0 +1,160 @@
...
+  gateway: kyma-system/kyma-gateway
+  host: httpbin
+  service:
+    name: httpbin
+    port: 8000
+  rules:
+    - path: /headers
+      methods: ["GET"]
+      externalAuthorizers:
+      - "oauth2-proxy" # Assuming that we will support an array
```
videlov, line 90: imo we should support single ext authoriser per path. if you have multiple APs I guess it will be treated with a logical `AND` which is not optimal.

videlov: we should think about if we want to restrict ext authorizer per `accessStrategy` do we want to mix it? what would be ext authorizer with `noop` AS?

triffer: > we should think about if we want to restrict ext authorizer per accessStrategy do we want to mix it? what would be ext authorizer with noop AS?

Later there will be only jwt and no_auth. For no_auth it should not be possible, but I see a use-case where you use JWTs for OAuth2 and what to make additional authorization decisions based on claims.

docs/user/technical-reference/external-authorizer.md
```diff
@@ -0,0 +1,129 @@
...
+    port: 8000
+  rules:
+    - path: /headers
+      methods: ["GET"]
+      mutators:
+        - ***
+      accessStrategy: # Validation: there needs to be one access strategy, and only one
+        extAuth:
+          name: oauth2-proxy # Validation: Check if there is that authorizer in Istio mesh config
+          restrictions:
```
triffer, line 114: Do we need this restrictions field? Maybe it's better to have it on the same level as in the jwt strategy.

barchw: Yeah, maybe it's better from API perspective
Just thought that this will make it easier to support this, as we can make a composition from the structure that will represent `extAuth`. 
Like this:

```go
type ExtAuthHandler struct{
  Restrictions JwtConfig
}
```

Though I think the same solution would be using unnamed field:
```go
type ExtAuthHandler struct{
  JwtConfig
}
```
I'm only not sure whether kubebuilder generator will be fine with that 😠

barchw: OK, I went through with deleting this field

PR #934, opened 2024-03-12
Wait for cluster role and service account before deploying oathkeeper

werdes72:
**Description**

- Wait for resources before deploying oathkeeper


barchw (commented):
LGTM, although I couldn't reproduce. @kolodziejczak could you take a look?

triffer:
LGTM
I'm not a big fan of this waiting, but we already had to do this. It would be better to get this to self-heal during the next reconciliation. I think the cleanest solution would be that the maester never reports Readiness.
Hopefully it's only a matter of time until we get rid of it.

triffer (approved):

PR #950, opened 2024-03-19
go 1.22

triffer:
**Description**

- Update go to 1.22
- Update controller-tools to 0.14.0 due to memory issue with go 1.22

**Related issues**


triffer:
`Pull Request / Run unit tests (pull_request)` fails because for code coverage check it will also run tests on main. Unfortunately, the controller-tools version on main doesn't work with go 1.22, that's why it fails.

kolodziejczak (approved):

Issue #870, opened 2024-02-13
Host link in APIRule in UI broken 


triffer:
**Description**

When you add an APIRule that has only a short "host" defined, e.g. `host: httpbin` the host in the APIRule UI is broken.
The host is incorrect in the list and details view.
![image](https://github.com/kyma-project/api-gateway/assets/11753933/c1e27bc5-8d36-4bc5-9e32-cdb6d37f647b)
![image](https://github.com/kyma-project/api-gateway/assets/11753933/526ec57b-a7a9-4a47-aff3-c8d4a658f008)

**Expected result**
The host link in the APIRule list and details is valid and doesn't contain `null`.

**Actual result**
The APIRule host link is broken.

**Steps to reproduce**

- Apply APIRule without full host name
```
kubectl apply -f - <<EOF
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: httpbin-rule
  namespace: default
  labels:
    app: httpbin
spec:
  gateway: kyma-system/kyma-gateway
  host: httpbin
  service:
    name: httpbin
    port: 8000
  rules:
    - path: /anything
      methods: [ "GET", "POST"]
      mutators: [ ]
      accessStrategies:
        - handler: no_auth
EOF
```
- Check APIRules list and details of the APIRule

**Troubleshooting**

PR #981, opened 2024-04-08
Fix wrong host in busola UI

werdes72:
**Description**

- Always display the configured value of Host and only add a link to the exposed domain.
- Change the preset format of APIRule gateway to `namespace/name`
- Update tests to work with redesigned Busola
- Run tests on chromium again

**Related issues**
#870


config/ui-extensions/apirules/details
```diff
@@ -11,8 +11,8 @@ header:
     source: 'status.APIRuleStatus.code ? status.APIRuleStatus.code : "UNKNOWN"'
     description: status.APIRuleStatus.desc
   - name: host
-    source: spec
-    widget: APIRuleHost
+    widget: ExternalLink
+    source: '$virtualServices().items[0].spec.hosts[0]'
```
barchw, line 15: This will only work when the VS is updated with the up to date host.
If, for example an update would cause the APIRule to go to `error` and not update the VS, it will be showing the host from the VS, not the actual one on APIRule

Simple scenario:
- create APIRule 1 with host a.example.com
- create APIRule 2 with host b.example.com
- update APIRule 2 with host a.example.com

The APIRule 2 will go into error, and the shown host will be `b.example.com`

triffer: While this is true, I think the only sane solution to have this link is to derive it from the VS, because deriving it from the APIRule and Gateway resources is error-prone and very complex. Especially since you can define the gateway in different format in the APIRule.

Also the whole UI must be reimagined for APIRule v1beta2, since we support multiple hosts.

Another solution could be to have the host field render the actual value and only if APIRule is in Ready state replace it with a link taken from the VS.
What do you think? @barchw

triffer: Since all our UI tests are e2e right now and I think it's okay. I don't think we'll change this in the next months, especially since to really decouple them a big change in how Busola processes UI supplied by modules would be required.

triffer:
Instead of always showing the link in the host field, I decided to use the field value instead, because I think it's a bad idea to add additional data to a field configured by the user, just for display. When the user then enters the edit mode, the link would turn into the simple host config again. I think it can be confusing for the user to see a different host than they actually configured on the APIRule.

The configured host will still be a copyable link, but a decision had to be made where to extract the link value from. Since the templating is limiting us a lot, the decision was made to use the host exposed in the Virtual Service. The Virtual Service will only reliably hold the correctly configured Host, if the APIRule is in OK state. That means for other states, the link will just fallback to the dummy value of the configured host value.
The reason for this is again limited possibilities in the templating. We cannot conditionally change the widget to use a Text widget, when the APIRule is not in OK state. 
Therefore the only other option would be to add an additional Text widget for the host, that is only visible when APIRule is not in OK state and hide the ExternalLink widget. While this can be done for the details view, this will not work for the list view, since the header of the table will always render both fields.

If we don't want to use the dummy link for the Host when APIRule is not in OK state, then in my opinion the only good solution is  to completely remove the link and always just display the value that is configured without a link.

kolodziejczak (approved):

PR #983, opened 2024-04-09
Bump golang in cicd test jobs

kolodziejczak:
**Description**

- Bump golang in cicd test jobs


triffer (changes requested):
We should use a similar approach as in the other workflows by using `go-version-file: 'go.mod'`.

triffer (approved):

Issue #821, opened 2024-01-18
API rule CORS values editor behaves in a strange way


ptesny:
[ the screenshots ](https://sap-btp.slack.com/archives/C01LGCBS196/p1705573106890689)

PR #1001, opened 2024-04-17
chore: Update maxAge documentation and placeholder

barchw:
**Description**

- Add placeholder value
- Update maxAge parameter documentation about the type of the field

**Related issues**
- https://github.com/kyma-project/api-gateway/issues/821


triffer (commented):
I wasn't aware that this is a duration. It might be a better UX to not have this as a duration but as a number in seconds, since the actually refers to seconds. Should we discuss this for v1beta2?

I think with that we can also easily improve validation in CRD, since right now it's possible to add a negative duration and this doesn't make sense at all, e.g. `-300s`. 

Besides this if we don't want to change the type of maxAge, we can think about setting an input info to have a better UX, e.g. `A duration string consist of decimal numbers, each with optional fraction and a unit suffix, such as "300ms" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h", "d", "w", "y".`

barchw:
> I wasn't aware that this is a duration. It might be a better UX to not have this as a duration but as a number in seconds, since the actually refers to seconds. Should we discuss this for v1beta2?
> 
> I think with that we can also easily improve validation in CRD, since right now it's possible to add a negative duration and this doesn't make sense at all, e.g. `-300s`.
> 
> Besides this if we don't want to change the type of maxAge, we can think about setting an input info to have a better UX, e.g. `A duration string consist of decimal numbers, each with optional fraction and a unit suffix, such as "300ms" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h", "d", "w", "y".`
> 
> What do you think?

I basically fully agree :D

triffer (approved):

docs/user/custom-resources/apirule/04-10-apirule-custom-resource.md
```diff
@@ -13,28 +13,28 @@ This table lists all parameters of APIRule CRD together with their descriptions:
...
+| **corsPolicy.allowCredentials**   | **NO**      | Specifies whether credentials are allowed in the **Access-Control-Allow-Credentials** CORS header.                                                                                                                                                                                                                                           |
+| **corsPolicy.exposeHeaders**      | **NO**      | Specifies headers exposed with the **Access-Control-Expose-Headers** CORS header.                                                                                                                                                                                                                                                            |
+| **corsPolicy.maxAge**             | **NO**      | Specifies the maximum age of CORS policy cache. The value is provided in the **Access-Control-Max-Age** CORS header. The value type is "duration", for example 200s.                                                                                                                                                                         |
```
nataliasitko, line 25:
```suggestion
| **corsPolicy.maxAge**             | **NO**      | Specifies the maximum age of CORS policy cache. The value is provided in the **Access-Control-Max-Age** CORS header. The value type is `duration`, for example, 200s.                                                                                                                                                                         |
```

barchw: The format is `1h200s`, so it should stay as it is.
> ParseDuration parses a duration string. A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h", "d", "w", "y".

nataliasitko: So let's also use code font for this

nataliasitko (approved):

PR #1018, opened 2024-04-25
Run ui tests on schedule

kolodziejczak:
**Description**

- Run ui tests on schedule


barchw:
>run on stage busola
This is running on production busola

.github/workflows/ui-tests.yaml
```diff
@@ -14,6 +14,8 @@ on:
     branches:
       - "main"
       - 'release-**'
+  schedule:
```
triffer, line 17: I don't know how the event structure in the GitHub API is, but are we sure, that `if: github.event.pull_request.draft == false` will not also apply for the schedule trigger and therefore trigger `run-ui-tests` and `run-ui-tests-periodic`?

.github/workflows/ui-tests.yaml
```diff
@@ -34,10 +36,33 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: Run tests
+        run: |
+          sudo echo "127.0.0.1 local.kyma.dev" | sudo tee -a /etc/hosts
+          wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | sudo bash
+          IMG=europe-docker.pkg.dev/kyma-project/dev/api-gateway-manager:PR-${{ github.event.number }} ./tests/ui/tests/scripts/k3d-ci-kyma-dashboard-integration.sh stage
```
triffer, line 62: I think we must not use the tag `PR-${{ github.event.number }}` for api-gateway-manager, but use the latest version from main.

triffer (changes requested):

.github/workflows/ui-tests.yaml
```diff
@@ -34,10 +36,33 @@ jobs:
...
+          IMG=europe-docker.pkg.dev/kyma-project/dev/api-gateway-manager:PR-${{ github.event.number }} ./tests/ui/tests/scripts/k3d-ci-kyma-dashboard-integration.sh prod
       - uses: actions/upload-artifact@v4
         if: always()
         name: Export Cypress output
         with:
           name: cypress
           path: tests/ui/tests/cypress/
+
+  run-ui-tests-periodic:
+    name: Run UI tests
```
triffer, line 48: We should have a different display name for this, since it can be confusing when it's in the same workflow. Also having it in the same workflow will now display it as skipped on every PR. Maybe it's better to add all jobs running on schedule in its own workflow.

.github/workflows/ui-tests.yaml
kolodziejczak: I've just separated two workflows, since in 2nd look it looks cleaner this way.

.github/workflows/ui-tests.yaml
kolodziejczak: I've just separated two workflows, since in 2nd look it looks cleaner this way. Also fixed those oversights.

kolodziejczak: Also seeing skipped job all the time on PR wouldn't be too nice.

.github/workflows/ui-tests.yaml
kolodziejczak: Right, I've just done that

.github/workflows/ui-tests-periodic.yaml
```diff
@@ -0,0 +1,47 @@
...
+        name: Export Cypress output
+        with:
+          name: cypress
+          path: tests/ui/tests/cypress/
+
+  slack_failed_notification:
+    name: Slack Notification
+    runs-on: ubuntu-latest
+    if: ${{ failure() }}
+    needs: [integration-tests, upgrade-tests, custom-domain-integration-gcp, custom-domain-integration-aws]
```
triffer, line 39: The jobs are not correct.

.github/workflows/ui-tests-periodic.yaml
```diff
@@ -0,0 +1,47 @@
+name: UI Tests Periodic
+
+on:
+  schedule:
+    - cron: '0 5 * * 1' # run every Monday at 5:00 AM
+
+permissions:
```
triffer, line 7: Why do we need this? I think read should be provided per default for the GitHub token in Actions.

.github/workflows/ui-tests.yaml
triffer: Ok, I think this might be an input for our team goal. It might be better to have one workflow file for all scheduled jobs.

triffer (approved):

nataliasitko (approved):

PR #1043, opened 2024-05-08
Adjust workflow to run proper action

kolodziejczak:
**Description**

- Adjust workflow to run proper action


tests/integration/main_test.go
```diff
@@ -164,3 +164,5 @@ func createDefaultContext(t *testing.T) context.Context {
 	ctx := testcontext.SetK8sClientInContext(context.Background(), client.GetK8sClient())
 	return testcontext.SetTestingInContext(ctx, t)
 }
+
+// test
```
werdes72, line 168: Is it needed?

kolodziejczak: only to trigger job to see if it works

kolodziejczak: job's green, trigger deleted

PR #1072, opened 2024-05-22
Remove apirule v1beta2 subresource statuses

barchw:
**Description**

- Remove apirule v1beta2 subresource statuses


apis/gateway/v1beta2/apirule_types.go
```diff
@@ -65,14 +65,9 @@ type Host string
...
-	LastProcessedTime    *metav1.Time           `json:"lastProcessedTime,omitempty"`
-	ObservedGeneration   int64                  `json:"observedGeneration,omitempty"`
-	APIRuleStatus        *APIRuleResourceStatus `json:"APIRuleStatus,omitempty"`
-	VirtualServiceStatus *APIRuleResourceStatus `json:"virtualServiceStatus,omitempty"`
-	// +optional
-	RequestAuthenticationStatus *APIRuleResourceStatus `json:"requestAuthenticationStatus,omitempty"`
-	// +optional
-	AuthorizationPolicyStatus *APIRuleResourceStatus `json:"authorizationPolicyStatus,omitempty"`
+	LastProcessedTime  *metav1.Time           `json:"lastProcessedTime,omitempty"`
+	ObservedGeneration int64                  `json:"observedGeneration,omitempty"`
```
triffer, line 69: Do we still need this? 🤔

apis/gateway/v1beta2/apirule_types.go
```diff
@@ -65,14 +65,9 @@ type Host string
...
+	APIRuleStatus      *APIRuleResourceStatus `json:"APIRuleStatus,omitempty"`
```
triffer, line 70: Should we rename this to Status to match what we have in other CRs?

triffer, line 70: The APIRuleResourceStatus has a little bit different naming than we use for other resources:
```go
type APIRuleResourceStatus struct {
	Code        StatusCode `json:"code,omitempty"`
	Description string     `json:"desc,omitempty"`
}
```
Should we change `Code` to `State` to match what we have in other CRs?

barchw: I was also wondering about moving it up and using the same thing that is used for module CRs.
Maybe this would be a better option?

apis/gateway/v1beta2/apirule_conversion_test.go
```diff
@@ -255,6 +255,49 @@ var _ = Describe("APIRule Conversion", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(Equal("either jwt is configured or noAuth must be set to true in a rule"))
 		})
+		It("should convert OK status from APIRuleStatus to v1beta2 Ready state", func() {
```
triffer, line 258:
```suggestion
		It("should convert v1beta2 Ready state to OK status from APIRuleStatus", func() {
```

triffer (approved):

Issue #1016, opened 2024-04-25
APIRule reconcile function refactoring


strekm:
**Description**

Introduction of APIRule v2alpha1 added complexity to main reconcile function that needs to be addressed before adding extAuth handler. 

ACs:
- [x] removed complexity from reconcile function
- [x] make implementation easy to extend with another handler
- [x] make reconcile function version aware
- [x] Create follow up issues

**Additional info**
- Configuration for immutable fields can be achieved by using CEL validation:
    ```go
    // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
    ```
    This might be useful to support conversion from v1beta2 to 1 without need for preserveUnknownFields.

**DoD:**
- [x] Provide unit ~~and integration tests.~~ -> integration tests in follow up, but also migration integration tests exist that cover the implementation.
- [x] Provide documentation.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
~~- [ ] If you changed the resource limits, explain why it was needed.~~
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
  => code coverage decreased from 61,1% to 59,2% but this is also due to open todos like for v2alpha1 validation, because some of the code already exists to match the interfaces, but the real implementation and tests will be done in a follow up.
~~- [ ] Add release notes.~~

**Attachments**
part of: #939 

**PR**
- https://github.com/kyma-project/api-gateway/pull/1140

**Follow ups**
- https://github.com/kyma-project/api-gateway/issues/1147
- https://github.com/kyma-project/api-gateway/issues/1148
- https://github.com/kyma-project/api-gateway/issues/1149
- https://github.com/kyma-project/api-gateway/issues/1151

PR #1140, opened 2024-07-02
Refactor APIRule reconciliation loop

barchw:
**Description**

- Separate the v2alpha1 loop from v1beta1

**Pre-Merge Checklist**

**Related issues**

- https://github.com/kyma-project/api-gateway/issues/1016


internal/processing/status/status.go
```diff
@@ -0,0 +1,45 @@
+package status
+
+import (
+	"github.com/kyma-project/api-gateway/apis/gateway/versions"
+	"github.com/kyma-project/api-gateway/internal/validation"
+)
+
+type ReconciliationStatusVisitor interface {
```
triffer, line 8: To avoid confusion I think we should not name it Visitor, since we don't have a "typical" visitor implementation.
```suggestion
type ReconciliationStatus interface {
```

internal/processing/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+	VisitStatus(status Status) error
```
triffer, line 9:
```suggestion
	UpdateStatus(status Status) error
```

internal/processing/status/v1beta1/v1beta1.go
```diff
@@ -0,0 +1,136 @@
+package v1beta1
```
triffer, line 1: Is it a benefit to have this in a separate v1beta1 package? Maybe it's better to move this and v2alpha1 status to the `status` package, especially since there is some duplicated code in these files that can be reused.

internal/validation/v1beta1/v1beta1.go
```diff
@@ -0,0 +1,332 @@
...
+	InjectionValidator        *validation.InjectionValidator
+	RulesValidator            rulesValidator
+	ServiceBlockList          map[string][]string
+	DomainAllowList           []string
+	HostBlockList             []string
+	DefaultDomainName         string
+}
+
+const (
+	Istio = iota
```
triffer, line 35: Looks like this is not used.

config/manager/kustomization.yaml
```diff
@@ -5,5 +5,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: api-gateway-controller
-  newTag: latest
+  newName: i566483/api-gateway-local
```
triffer, line 8: We need to revert this.

controllers/gateway/reconcile_result.go
```diff
@@ -40,7 +40,6 @@ func (r *APIRuleReconciler) updateStatus(ctx context.Context, api *gatewayv1beta
 	api.Status.ObservedGeneration = api.Generation
 	api.Status.LastProcessedTime = &v1.Time{Time: time.Now()}
 
-	// TODO: api status as dependency injection
```
barchw: This todo is still relevant for the status follow up

triffer: Can we add more information to the follow up task instead of putting inline comments? Also the comment doesn't give much information. I sharing the idea in the follow up task what should be changed in what way will be more helpful.

barchw: OK, agree :)

triffer (dismissed):

triffer (approved):

docs/contributor/04-10-technical-design.md
```diff
@@ -32,13 +32,23 @@ APIRule Controller has a conditional dependency to APIGateway Controller in term
...
-You can use the [API Gateway Operator paramteres](../user/technical-reference/05-00-api-gateway-operator-parameters.md) to adjust this interval.
+You can use the [API Gateway Operator parameters](../user/technical-reference/05-00-api-gateway-operator-parameters.md) to adjust this interval.
 In the event of a failure during the reconciliation, APIRule Controller performs the reconciliation again after one minute.
 
 The following diagram illustrates the reconciliation process of APIRule and the created resources:
 
 ![APIRule CR Reconciliation](../assets/api-rule-reconciliation-sequence.svg)
 
+#### Reconciliation Processors
+The APIRule reconciliation supports different processors that are responsible for validation, status handling and creating, updating and deleting the resources in the cluster.
```
nataliasitko, line 43:
```suggestion
The APIRule reconciliation supports different processors that are responsible for validation and status handling as well as creating, updating, and deleting the resources in the cluster. 
```

triffer (approved):

nataliasitko (approved):

PR #1242, opened 2024-08-09
cicd: bump k8s versions for int tests

kolodziejczak:
**Description**

- bump k8s versions for int tests
- drop external gh action responsible for provisioning k3d
- provide own gh action for provisioning k3d

### DIFF
--- a/.github/actions/integration-tests/action.yaml
+++ b/.github/actions/integration-tests/action.yaml
@@ -47,9 +47,9 @@ runs:
         IMG: ${{ inputs.manager_image }}
       run: |
         if [ "${{ github.event_name }}" == "pull_request" ]; then
-          k3d image import ${{ inputs.manager_image }} -c test-cluster-1
+          k3d image import ${{ inputs.manager_image }} -c k3s-default
         fi
-        kubectl config use-context k3d-test-cluster-1
+        kubectl config use-context k3s-default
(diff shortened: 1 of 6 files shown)


barchw (approved):

barchw (approved):

.github/actions/integration-tests/action.yaml
```diff
@@ -28,15 +28,15 @@ runs:
...
-        cluster-name: "test-cluster-1"
-        args: >-
-          --agents 0
-          --port 80:80@loadbalancer
-          --port 443:443@loadbalancer
-          --k3s-arg "--disable=traefik@server:0"
+    - name: Create Cluster
+      shell: bash
+      run: |
+        curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```
barchw, line 34: I would maybe move installation of k3d to a different step. WDYT?

kolodziejczak: I was thinking about it, but since it's pretty short I don't have any hard opinion on that. If you consider it cleaner, we can do that.

barchw: I would do it in case we want to use it somewhere else, just for description of what it is doing

barchw (approved):

barchw (approved):

PR #1259, opened 2024-08-20
Fix conversion for v2alpha1 APIRule with JWT authorizations only

videlov:
… configuration with authorizations only

**Description**

- Fix conversion and validation for when v2alpha1 APIRule specifies JWT configuration with authorizations only

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/internal/validation/v2alpha1/gateway.go
+++ b/internal/validation/v2alpha1/gateway.go
@@ -2,6 +2,7 @@ package v2alpha1
 
 import (
 	"fmt"
+
 	gatewayv2alpha1 "github.com/kyma-project/api-gateway/apis/gateway/v2alpha1"
 	"github.com/kyma-project/api-gateway/internal/validation"
 	networkingv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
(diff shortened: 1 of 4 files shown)


internal/validation/v2alpha1/gateway.go
```diff
@@ -2,20 +2,24 @@ package v2alpha1
...
 func validateGateway(parentAttributePath string, gwList networkingv1beta1.GatewayList, apiRule *gatewayv2alpha1.APIRule) []validation.Failure {
 	var failures []validation.Failure
-	gatewayName := *apiRule.Spec.Gateway
 
-	if !gatewayExists(gwList, gatewayName) {
-		failures = append(failures, validation.Failure{
-			AttributePath: parentAttributePath + ".gateway",
-			Message:       "Gateway not found",
-		})
+	if apiRule.Spec.Gateway != nil {
```
barchw, line 14: It should already fail here, not return empty

videlov: let me check if this is applicable, for APIRules that can not be converted (e.g. empty spec).

videlov: looks good, added additional/more general validation step for empty spec

barchw (approved):

PR #1279, opened 2024-09-03
Update v2alpha1 APIRule rule path validation

triffer:
**Description**

- Update v2alpha1 APIRule rule path validation to allow wildcard (*) only if no other characters exist. This behaviour is documented already and also described in the ADR.
- Update integration tests to use `*` instead of `/*` or `/.*`

**Pre-Merge Checklist**

**Related issues**


triffer:
Closing since it was only used for running integration tests

Issue #1282, opened 2024-09-04
APIRule v2alpha1 path validation and behaviour is wrong


strekm:
**Description**

According to [proposal](https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0007-apirule-v2alpha1-api-proposal.md) path can be either exact match or wildcard `*`. 

APIRule v2alpha1 accepts `./*` as valid path. Calling workload results in 404.  

**Expected result**

Accepted path should be either exact match or `*`. Workload does not return 404.

**Actual result**

`./*` is accepted as valid. Workload returns 404

**Steps to reproduce**

```
export NAMESPACE=test
export GATEWAY=kyma-system/kyma-gateway
export DOMAIN_TO_EXPOSE_WORKLOADS={KYMA_DOMAIN}

cat <<EOF | kubectl apply -f -
apiVersion: gateway.kyma-project.io/v2alpha1
kind: APIRule
metadata:
  name: httpbin
  namespace: $NAMESPACE
spec:
  hosts: 
    - httpbin.$DOMAIN_TO_EXPOSE_WORKLOADS
  service:
    name: httpbin
    namespace: $NAMESPACE
    port: 8000
  gateway: $GATEWAY
  rules:
    - path: ./*
      methods: ["GET"]
      noAuth: true
EOF
```

**Troubleshooting**
Tests for validation and first implementation to fix this can be found in the PR https://github.com/kyma-project/api-gateway/pull/1285. 

**PR**
- https://github.com/kyma-project/api-gateway/pull/1285

strekm:
ok so `./*` was typo on my side. either way it should not be accepted as valid. only `*` or exact match should be valid

triffer:
It was agreed that we will allow `/*` instead of `*` in the rule path to expose all paths. The ADR should also be updated with this topic.

PR #1285, opened 2024-09-06
Update v2alpha1 APIRule rule path validation

triffer:
**Description**

This PR introduces restrictions to using `*` in APIRule path.
Instead of allowing `/.*` the path will only allows `/*` if a rule should apply to all paths.
In the [ADR](https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0007-apirule-v2alpha1-api-proposal.md) it's also mentioned that a configuration of `*` `Should be warning user that it is not recommended, as it applies to all paths`. This will not be handled in this PR, but should be handled once there is support for APIRule v2alpha1 in the UI.

- Update allowed wildcard from `*` to `/*` in the ADR and documentation 
- Update v2alpha1 APIRule rule path validation to allow `*` in the format of `/*`.
- Simplify code for v2alpha1 authorization policy creation, because `*` is supported in the path.
- Update integration tests to use `/*` instead of `/.*`
- Remove unnecessary integration test

**Pre-Merge Checklist**

**Related issues**

- https://github.com/kyma-project/api-gateway/issues/1282


barchw (approved):
Technical side looks good.

Should (or did we already) document that customers need to switch their `/.*` paths to `/*` as part of migration?

triffer:
> Technical side looks good.
> 
> Should (or did we already) document that customers need to switch their `/.*` paths to `/*` as part of migration?

Good point, I'll check how we want to communicate this.

nataliasitko (approved):

PR #1286, opened 2024-09-06
feat: replace getting converted resource in reconcile

Ressetkk:
/kind feature
/area api-gateway

APIRule v2alpha1 was retrieved from APIServer. This introduced unnecessary dependency to conversion webhook. So instead of asking APIServer to get resource, call ConvertTo() and ConvertFrom() directly to convert v1beta1 resource into v2alpha1.

Additionally include error in doneReconcileErrorRequeue(), because it should return an errored reconcile result.

This resolves multiple instances of error
```
2024/09/06 14:02:53 http: TLS handshake error from 10.244.0.1:57417: EOF
```

**Related issues**

#1148

### DIFF
--- a/controllers/gateway/apirule_controller.go
+++ b/controllers/gateway/apirule_controller.go
@@ -79,7 +79,7 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	defaultDomainName, err := default_domain.GetDefaultDomainFromKymaGateway(ctx, r.Client)
 	if err != nil && default_domain.HandleDefaultDomainError(l, err) {
-		return doneReconcileErrorRequeue(errorReconciliationPeriod)
+		return doneReconcileErrorRequeue(err, errorReconciliationPeriod)
 	}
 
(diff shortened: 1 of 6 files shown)


controllers/gateway/apirule_controller.go
```diff
@@ -202,28 +201,40 @@ func (r *APIRuleReconciler) reconcileV2alpha1APIRule(ctx context.Context, req ct
...
+	}
+	return r.convertAndUpdateStatus(ctx, l, rule)
+}
+
+// convertAndUpdateStatus is a small helper function that converts APIRule
+// resource from convertible v2alpha1 to hub version v2alpha1
+func (r *APIRuleReconciler) convertAndUpdateStatus(ctx context.Context, l logr.Logger,
+	rule gatewayv2alpha1.APIRule) (ctrl.Result, error) {
+	l.Info("Converting APIRule v2alpha1 to v1beta1")
+	n := gatewayv1beta1.APIRule{}
```
barchw, line 226: I think `n` might be a little bit to short of a name for `APIRule` ;)

barchw (changes requested):
One small nitpick, otherwise LGTM
Thanks for following up with this

Issue #1231, opened 2024-08-07
apigateway.operator.kyma-project.io/v1alpha1 has a hard requirement for CRD dnsentries.dns.gardener.cloud is not present in v2.5.0?


tehret77:
**Description**

Installed the latest version 2.5.0 of the APIGateway manager module on a Azure Kubernetes cluster (Kyma OpenSource).
Afterward applyied the default CR for a gateway.
This gateway ras the status Error with the message:
```status:
  conditions:
  - lastTransitionTime: "2024-08-07T13:52:13Z"
    message: Module dependencies missing
    reason: DependenciesMissing
    status: "False"
    type: Ready
  description: CRD dnsentries.dns.gardener.cloud is not present. Make sure to install
    required dependencies for the component
  state: Error 
  ```

**Expected result**

As in version 2.4.0 there is no hard dependency to any Gardener CRD

**Actual result**

There is a hard requirement.

**Steps to reproduce**

Install the module on a blank Kubernetes cluster (non Gardener)

**Troubleshooting**

**Comment by tehret77** [comment]
I think, I found the issue.
The application-connector module require the configmap kube-system/shoot-info for some configuration with compass agent. Once this configmap is present during initial configuration or upgrade the info from that configmaps is used and require the dependency to the CRD dnsentries.dns.gardener.cloud.

**Comment by Ressetkk** [comment]
I have found the part of the code that handles it. https://github.com/kyma-project/api-gateway/blob/a100420678f106b00440bb3832290a0412465746/internal/reconciliations/gardener.go#L28 If the cluster contains the shoot-info config map, it assumes that controller is working on Gardener, even if you create this resource manually. I think we need to find other solution for expecting gardener cluster.

**Comment by mluk-sap** [comment]
Discussed today during the refinement.

The Product Owner vetoed the idea of changing the API to have an explicit attribute to control the behavior. The recommendation is to implement an implicit check (Gardener CRDs existence). This is a common practice in other Kyma modules and in the Istio.

**Comment by mluk-sap** [comment]
Steps to reproduce:

```
k3d cluster create kyma --api-port 127.0.0.1:6443

kubectl create ns kyma-system

kubectl create -n kube-system cm shoot-info

kubectl label namespace kyma-system istio-injection=enabled --overwrite
kubectl apply -f https://github.com/kyma-project/istio/releases/latest/download/istio-manager.yaml
[...]

PR #1341, opened 2024-10-02
feat: change logic of handling Gardener cluster

Ressetkk:
/kind feature
/area api-gateway

Remove hard dependency to Gardener resources and handle them on-demand if they are present.
Introduce fallback mechaninsm if the resources expected on gardener cluster are not found.
Still leave hard dependency to resources needed generally by the manager.
Update tests to move the logic of dependency checking in reconcileKymaGateway (just adding required CRDs for fakes)
Remove injecting Gardener CRDs where they aren't added into envtest runtime schema
Create hack/crds directory containing all CRDs used in the tests

#1231


triffer (changes requested):
I implemented two tests that are verifying the scenario reported by the user and they are failing. I will commit those tests and we should make those tests green. Also the steps to reproduce mentioned [here](https://github.com/kyma-project/api-gateway/issues/1231#issuecomment-2387998541) still fail. This should also work once the tests are green.

triffer (approved):

PR #1354, opened 2024-10-07
feat: show logs as json

Ressetkk:
/kind feature
/area api-gateway

Move zap Development option to controller flag.
This enables projection of pure JSON-based logs with optional possibility to switch back to old view.

after the change logs look like this:
```
{"level":"info","ts":"2024-10-07T14:36:48+02:00","logger":"setup","msg":"Reading certficate secret","namespace":"kyma-system","name":"api-gateway-webhook-certificate"}
{"level":"info","ts":"2024-10-07T14:36:48+02:00","logger":"setup","msg":"Certificate validity","validFrom":"2024-09-06T09:19:15Z","validUntil":"2024-12-05T09:19:15Z"}
{"level":"info","ts":"2024-10-07T14:36:48+02:00","logger":"controller-runtime.builder","msg":"skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called","GVK":"gateway.kyma-project.io/v1beta1, Kind=APIRule"}
{"level":"info","ts":"2024-10-07T14:36:48+02:00","logger":"controller-runtime.builder","msg":"skip registering a validating webhook, object does not implement admission.Validator or WithValidator wasn't called","GVK":"gateway.kyma-project.io/v1beta1, Kind=APIRule"}
```
Instead of this:
```
2024-10-07T14:45:29+02:00       INFO    setup   Reading certficate secret       {"namespace": "kyma-system", "name": "api-gateway-webhook-certificate"}
2024-10-07T14:45:29+02:00       INFO    setup   Certificate validity    {"validFrom": "2024-09-06T09:19:15Z", "validUntil": "2024-12-05T09:19:15Z"}
2024-10-07T14:45:29+02:00       INFO    controller-runtime.builder      skip registering a mutating webhook, object does not implement admission.Defaulter or WithDefaulter wasn't called       {"GVK": "gateway.kyma-project.io/v1beta1, Kind=APIRule"}
2024-10-07T14:45:29+02:00       INFO    controller-runtime.builder      skip registering a validating webhook, object does not implement admission.Validator or WithValidator wasn't called     {"GVK": "gateway.kyma-project.io/v1beta1, Kind=APIRule"}
```

With falling back to old projection with an application flag `--log-devel`


kolodziejczak:
Do we want to automate this new flag insertion for the dev run?

kolodziejczak (commented):
thanks for catching that, should we make sure that it's not a breaking change for any logs gathering systems?

triffer:
I think it's a good idea to allow changing log format to json, but I think the name of the flag can be improved. I think it's rather not a development log, since we also don't use any structured logging in our prod systems. 😬 Something like `logJson` with a default of `true` might be easier to understand.
Additionally, I'd not use json log in our environments, since it makes the log harder to read and unfortunately, we don't have any logging backend, yet.

Ressetkk:
@triffer I see no clear standards across modules - warden and btp-operator use hardcoded Development mode set to true, but KLM doesn't. Looks like this could be improved on global scale.

triffer:
> @triffer I see no clear standards across modules - warden and btp-operator use hardcoded Development mode set to true, but KLM doesn't. IIRC the only standard that got set was to use zap as logger across modules. The rest was just copy-paste. Looks like this could be improved on global scale. IDK why SRE didn't start a riot by then...

I agree with you, that not having structured central logging for modules is very bad and I mentioned that in the past. I feel like SRE probably uses human readable format only, because as far as I know they don't use any log collector/log backend for modules.

PR #1375, opened 2024-10-11
fix: Check for Gateway existence after requesting v2alpha1 APIRule

barchw:
**Description**

- Check for Gateway existence after requesting v2alpha1 APIRule

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/controllers/gateway/apirule_controller.go
+++ b/controllers/gateway/apirule_controller.go
@@ -185,12 +185,13 @@ func (r *APIRuleReconciler) reconcileV2Alpha1APIRule(ctx context.Context, l logr
 		return doneReconcileErrorRequeue(err, r.OnErrorReconcilePeriod)
 	}
 
-	gateway, convertStatus, err := discoverGateway(r.Client, ctx, l, &rule)
-	if convertStatus {
-		return r.convertAndUpdateStatus(ctx, l, rule)
-	}
+	gateway, err := discoverGateway(r.Client, ctx, l, &rule)
 	if err != nil {
 		return doneReconcileErrorRequeue(err, r.OnErrorReconcilePeriod)
+	}
+
+	if gateway == nil {
+		return r.convertAndUpdateStatus(ctx, l, rule)
 	}
 
 	cmd := r.getV2Alpha1Reconciliation(&apiRule, &rule, gateway, migrate, &l)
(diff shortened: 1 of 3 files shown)


controllers/gateway/apirule_controller.go
```diff
@@ -285,6 +269,50 @@ func handleDependenciesError(name string, err error) controllers.Status {
...
 }
 
+func discoverGateway(client client.Client, ctx context.Context, l logr.Logger, rule *gatewayv2alpha1.APIRule) (*networkingv1beta1.Gateway, bool, error) {
+	if rule.Spec.Gateway == nil {
+		return nil, false, fmt.Errorf("expected Gateway to be set")
+	}
+
+	// The regex pattern is the exact same as the one used in the APIRule validation
+	// Unfortunately usage of a constant is not possible here, as the Kubebuilder CRD validation interface is comment based
+	match, err := regexp.MatchString(`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?/([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)$`, *rule.Spec.Gateway)
```
videlov, line 279: The original issue is that APIRule `spec.Gateway` is empty, correct? I would not duplicate regex matching here, but just check for the case when empty spec is produced (e.g. `spec.Gateway` is empty) on conversion, because of not-supported case.

controllers/gateway/apirule_controller.go
```diff
@@ -285,6 +269,50 @@ func handleDependenciesError(name string, err error) controllers.Status {
...
+			{
+				AttributePath: "spec.gateway",
+				Message:       "Could not get specified Gateway",
+			},
+		})
+		if err := s.UpdateStatus(&rule.Status); err != nil {
+			l.Error(err, "Error updating APIRule status")
+			return nil, false, err
+		}
+		return nil, true, err
```
videlov, line 310: can we get rid of `convertStatus` return parameter? having 3 kind of increase complexity, reduce readability and can be case-matched by other two return parameters, e.g. `true` if `return nil, nil` happens and `false` otherwise.

controllers/gateway/discover_gateway_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Expect(gotGateway.Namespace).To(Equal(gatewayNamespace))
+	},
+		Entry("should return gateway when it exists", &v2alpha1.APIRule{
+			Spec: v2alpha1.APIRuleSpec{
+				Gateway: ptr.To(fmt.Sprintf("%s/%s", gatewayNamespace, gatewayName)),
+			},
+		}, true, false),
+		Entry("should return error when gateway does not exist", &v2alpha1.APIRule{
+			Spec: v2alpha1.APIRuleSpec{
+				Gateway: nil,
```
videlov, line 71: perhaps also test when Gateway empty string? not sure if this is a valid case. if empty spec is Gateway=nil, this is enough.

controllers/gateway/discover_gateway_test.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+		}, false, true),
+		Entry("should return error when gateway is not in namespacedName format", &v2alpha1.APIRule{
+			Spec: v2alpha1.APIRuleSpec{
+				Gateway: ptr.To(fmt.Sprintf("%s.%s.svc.cluster.local", gatewayName, gatewayNamespace)),
```
videlov, line 76: this is not needed if you decide to drop the reg-ex match.

tests/integration/pkg/helpers/api_rule.go
```diff
@@ -124,12 +125,12 @@ func ApplyApiRuleV2Alpha1ExpectError(toExecute RetryableApiRule, onRetry Retryab
 				return err
 			}
 			if apiStatus.Status.State != "Error" {
-				log.Println("APIRule status not Error: " + apiStatus.Status.Description)
-				return errors.New("APIRule status not Error: " + apiStatus.Status.Description)
+				log.Printf("expected but APIRule status to be Error, got %s with desc: %s", apiStatus.Status.State, apiStatus.Status.Description)
```
videlov, line 128: why is this, perhaps not needed for this PR?

videlov (changes requested):

controllers/gateway/apirule_controller.go
barchw: I found another bug unfortunately, you can edit the gateway format in v1beta1 CRD version.

This snippet assumes that the format is always OK, meaning that array out of bound error will happen on such occasion.
```go
	gatewayName := strings.Split(*rule.Spec.Gateway, "/")
 	gatewayNN := types.NamespacedName{
 		Namespace: gatewayName[0],
 		Name:      gatewayName[1],
 	}
```

videlov: okay, so we can not then skip on matching again, perhaps just add comment (to not forget) to remove the double reg-ex definition when re get rid of v1beta1 and conversion.

controllers/gateway/discover_gateway_test.go
videlov: we need the reg-ex, for the v1beta1 old format gateway + conversion case, ending up with v2alpha gateway in old format. so this is fine.

videlov (approved):

PR #1379, opened 2024-10-14
Example for sub-path wildcarding for v2alpha1

videlov:
**Description**

- Example for sub-path wildcarding for v2alpha1

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/apis/gateway/v2alpha1/apirule_types.go
+++ b/apis/gateway/v2alpha1/apirule_types.go
@@ -119,8 +119,8 @@ type Service struct {
 // Rule .
 // +kubebuilder:validation:XValidation:rule="((has(self.extAuth)?1:0)+(has(self.jwt)?1:0)+((has(self.noAuth)&&self.noAuth==true)?1:0))==1",message="One of the following fields must be set: noAuth, jwt, extAuth"
 type Rule struct {
-	// Specifies the path of the exposed service, allowing '/*' to match all paths and '/a-path' to match a specific path.
-	// +kubebuilder:validation:Pattern=^(\/\*|\/[0-9a-zA-Z/()?!\\_-]+)$
+	// Specifies the path of the exposed service, allowing '/*' to match all paths, '/a-path' to match a specific path or match multiple sub-paths having '*' wildcards.
+	// +kubebuilder:validation:Pattern=`^(\/[^\?\s#]+)(?:[\?#].*)?$`
 	Path string `json:"path"`
 	// Describes the service to expose. Overwrites the **spec** level service if defined.
 	// +optional
--- a/config/crd/bases/gateway.kyma-project.io_apirules.yaml
+++ b/config/crd/bases/gateway.kyma-project.io_apirules.yaml
@@ -522,9 +522,9 @@ spec:
                       type: boolean
                     path:
                       description: Specifies the path of the exposed service, allowing
-                        '/*' to match all paths and '/a-path' to match a specific
-                        path.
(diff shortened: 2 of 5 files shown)


triffer:
The proposed regex `(\/[^\?\s#]+)(?:[\?#].*)?` should not be used, because it also allows `/something/*/anything`
and this is currently not supported by Istios.
We only want to support a `*` as prefix or postfix, but not in the middle of a path.

We also need to answer the general question if we can support a path configuration like `/callback/*` only by updating the validation or if more effort is needed.

I did some testing to see how VS and AP can be configured to get more details about this.

## Testing VS

### v2alpha1
When creating APIRule like this:
```yaml
apiVersion: gateway.kyma-project.io/v2alpha1
kind: APIRule
metadata:
  name: httpbin-rule
  namespace: default
  labels:
    app: httpbin
spec:
  gateway: kyma-system/kyma-gateway
  hosts:
    - httpbin
  service:
    name: httpbin
    port: 8000
  rules:
    - path: /*
      methods: [ "GET"]
      noAuth: true
```

The VS is created with `prefix: /`.
The AP is created with `to.operation.paths: ["/*"]`.
This will return status 200 on requests to httpbin endpoint `/anything` and `/anything/allow`.

**Changing VS to `prefix: /anything/*` and having AP set to `to.operation.paths: ["/*"]`**
Requests to `/anything` and `/anything/allow` fail with 404.

**Changing VS to `regex: /anything/*` and having AP set to `to.operation.paths: ["/*"]`**
Requests to `/anything` return 200 and requests to `/anything/allow` fail with 404.

**Changing VS to `regex: /anything/.*` and having AP set to `to.operation.paths: ["/*"]`**
Requests to `/anything` return 404 and requests to `/anything/allow` fail with 200.

**Changing VS to `regex: "*/allow"` and having AP set to `to.operation.paths: ["/*"]`**
Blocked by admission webhook:
```
configuration is invalid: "uri": error parsing regexp: missing argument to repetition operator: `*`; Istio uses RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax)
```

**Changing VS to `regex: ".*/allow"` and having AP set to `to.operation.paths: ["/*"]`**
Requests to `/anything` return 404 and requests to `/anything/allow` return 200.

**Changing AP to `to.operation.paths: ["/anything/*"]` and having VS set to `prefix: /`**
Requests to `/anything` and `/anything/allow` return 200.

**Changing AP to `to.operation.paths: ["*/allow"]` and having VS set to `prefix: /`**
Requests to `/anything` return 403 and requests to `/anything/allow` return 200.

### v1beta1
When creating APIRule like this:
```yaml
apiVersion: gateway.kyma-project.io/v1beta1
kind: APIRule
metadata:
  name: httpbin-rule
  namespace: default
  labels:
    app: httpbin
spec:
  gateway: kyma-system/kyma-gateway
  host: httpbin
  service:
    name: httpbin
    port: 8000
  rules:
    - path: /anything/*
      methods: [ "GET"]
      mutators: [ ]
      accessStrategies:
        - handler: no_auth
```

The VS is created with `regex: /anything/*` and therefore requests to `/anything` return 200 and requests to `/anything/allow` return 404.

## User scenario
The user mentioned the configuration of the paths  `/subscription_callback/*`  and `/callback/*`.
In this case we need to implement additional logic to handle the conversion from `/callback/*` to `/callback/.*` in the Virtual Service.

Additionally, it will be required to adapt the logic for overlapping paths, e.g. if another rule is configured for the path `/callback/something`.

Issue #1397, opened 2024-10-23
Support {*} and {**} operators in APIRule v2alpha1 path


strekm:
**Description**

Based on outcome of POC and requirements from https://github.com/kyma-project/api-gateway/issues/1297 extends APIRule v2alpha1 path to support `{*}` and `{**}`. `{*}` and `{**}` behaviour is matching the one described in AP. It should be possible to use mentioned operators accordingly as described in AP docu. `/*` is considered secial case in APIRule and it represents wildcard match.

Supported validation examples:
- `/headers`
- `/status/{*}`
- `/anything/{**}/hello`
- `/*`

Not supported:
- `/a/*` -> this was not working in v1beta1, so there is no reason to support this use case.

For more information see the `paths` field description in [Istio documentation](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Operation).

TODOs:
- [x] Extend CRD validation
- [x] Improve path validation
- [x] Integration tests for different scenarios
- [x] Check VS and AP builders if they need adapting
- [x] Report issue to Istio about when `{**}` is used twice --> email sent

ACs:
- [x] it is possible to use `{*}` and `{**}` in path 
- [x] `{*}` and `{**}` behaviour is matching Istio AuthorizationPolicy behaviour
- [x] `/*` matches anything
- [x] path is validated including above rules
- [x] path is documented 

PRs:
- https://github.com/kyma-project/api-gateway/pull/1411

**Reasons**
Provide convenience for users exposing their workloads

**DoD:**
- [x] Provide unit and integration tests.
- [x] Provide documentation.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed.
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [x] Add release notes.

**Attachments**
https://istio.io/latest/docs/reference/config/security/authorization-policy/#Operation
fixes: https://github.com/kyma-project/api-gateway/issues/1297

PR #1411, opened 2024-10-29
Introduce support for the path operators `{*}` and `{**}`

barchw:
**Description**

- Add support for {*} and {**}

**Pre-Merge Checklist**

**Related issues**

- https://github.com/kyma-project/api-gateway/issues/1397


internal/validation/v2alpha1/rules.go
```diff
@@ -48,6 +49,7 @@ func validateRules(ctx context.Context, client client.Client, parentAttributePat
 			problems = append(problems, extAuthFailures...)
 		}
 
+		problems = append(problems, validatePath(ruleAttributePath, rule.Path)...)
```
Ressetkk, line 52: If you return slice of failures with only one failure, refactor it to append only one failure

internal/validation/v2alpha1/rules.go
```diff
@@ -56,6 +58,24 @@ func validateRules(ctx context.Context, client client.Client, parentAttributePat
 	return problems
 }
 
+func validatePath(validationPath string, rulePath string) (problems []validation.Failure) {
+	problems = append(problems, validateEnvoyTemplate(validationPath+".path", rulePath)...)
+	return problems
+}
+
+func validateEnvoyTemplate(s string, path string) []validation.Failure {
+	segments := strings.Split(strings.TrimLeft(path, "/"), "/")
```
Ressetkk, line 67: You can firstly check if you path has over 1 `{**}`. If it does, you can fail early.
```go
if strings.Count("{**}", path) > 1 {
    return []validation.Failure{}
}
```

internal/validation/v2alpha1/rules.go
```diff
@@ -56,6 +58,24 @@ func validateRules(ctx context.Context, client client.Client, parentAttributePat
...
+	doubleAsteriskExists := false
+	for _, segment := range segments {
```
Ressetkk, line 69: If you iterate over the slice of segments, only last operator is important to you to check if it's glob match operator. If your path has only one count of `{**}` and If last operator is `{**}`, then it's a valid path. See https://go.dev/play/p/1N_Uk-Llm-j

Ressetkk (changes requested):

internal/validation/v2alpha1/rules.go
barchw: This is easier to handle as the function does not need to return a pointer to validation.Failure, as slices are pointers by default. In that context what would you prefer: return slice, or return a pointer and check if the return is nil?

internal/validation/v2alpha1/rules.go
```diff
@@ -56,6 +58,24 @@ func validateRules(ctx context.Context, client client.Client, parentAttributePat
...
+		if segment == "{**}" && !doubleAsteriskExists {
+			doubleAsteriskExists = true
+		} else if (segment == "{*}" || segment == "{**}") && doubleAsteriskExists {
+			return []validation.Failure{{AttributePath: s, Message: fmt.Sprintf("Operator %s was used after operator {**}. The {**} operator must be the last in the path.", segment)}}
```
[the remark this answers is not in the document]
barchw: Problem is that in case you do not return a slice, you must always return value. This does not work with how we handle the failures, as the slice length will increase

https://go.dev/play/p/VMYDiP3KY8f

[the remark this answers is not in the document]
Ressetkk: But you can just return a single `validation.Failure` and append only one element in L62
```
problems = append(problems, validatePath(ruleAttributePath, rule.Path))
```

internal/validation/v2alpha1/rules.go
```diff
@@ -56,6 +58,45 @@ func validateRules(ctx context.Context, client client.Client, parentAttributePat
...
 }
 
+func validatePath(validationPath string, rulePath string) (problems []validation.Failure) {
+	problems = append(problems, validateEnvoyTemplate(validationPath+".path", rulePath)...)
+	return problems
+}
+
+func validateEnvoyTemplate(validationPath string, path string) []validation.Failure {
+	if strings.Count(path, "{**}") == 0 {
+		return []validation.Failure{}
```
barchw, line 68:
```suggestion
		return nil
```

Ressetkk (dismissed):

docs/release-notes/2.9.0.md
```diff
@@ -0,0 +1,3 @@
+## New Features
+
+- Introduce support for path operators `{*}` and `{**}`[#1397](https://github.com/kyma-project/api-gateway/issues/1397)
```
nataliasitko, line 3:
```suggestion
- Introduce support for the path operators `{*}` and `{**}`[#1397](https://github.com/kyma-project/api-gateway/issues/1397)
```

docs/user/custom-resources/apirule/v2alpha1/04-10-apirule-custom-resource.md
```diff
@@ -16,44 +16,44 @@ This table lists all parameters of APIRule `v2alpha1` CRD together with their de
...
+| **rules.path**                                   | **YES**  | Specifies the path on which the service is exposed. Supported configurations are: Exact path (e.g. `/abc`) - matches the specified path exactly; Usage of the `{*}` operator (e.g. `/foo/{*}` or `/foo/{*}/bar`) - match any request that matches the pattern with exactly one path segment in the operator's place; Usage of the `{**}` operator (e.g. `/foo/{**}` or `/foo/{**}/bar`), `{**}` must be used as the last operator in the path - match any request that matches the pattern with zero or more path segments in the operator's place; Wildcard path `/*` - matches all paths. Equivalent to `/{**}` path. | Value can contain operators `{*}` and/or `{**}`. Can also be a wildcard match `/*`                                    |
```
nataliasitko, line 36:
```suggestion
| **rules.path**                                   | **YES**  | Specifies the path on which the service is exposed. The supported configurations are:<ul><li>Exact path (e.g. `/abc`) - matches the specified path exactly.</li><li>Usage of the `{*}` operator (e.g. `/foo/{*}` or `/foo/{*}/bar`) - matches any request that matches the pattern with exactly one path segment in the operator's place.</li><li>Usage of the `{**}` operator (e.g. `/foo/{**}` or `/foo/{**}/bar`) - matches any request that matches the pattern with zero or more path segments in the operator's place. `{**}` must be the last operator in the path.</li><li>Wildcard path `/*` - matches all paths. It's equivalent to the `/{**}` path.</ul>
| The value might contain operators `{*}` and/or `{**}`. It can also be a wildcard `/*`.                                    |
```

nataliasitko (approved):

PR #1415, opened 2024-10-30
hack: consolidate pull-request status into one job

Ressetkk:
/kind chore
/area ci

Modify usage of tj-changed files in specific files. Any kind of status other than failure is treated as `success`.
Introduce pull-request-status job that returns `success`, if all jobs return success.
This PR moves the usage of changed files filter into respective files, making the jobs in `pull-request.yaml` always reported. If any of the jobs/steps fail, the pull-request-status job will result in a failure.
This job can be used as required status check

After this PR gets merged, new status checks will have to be set:
- pull-request-status

References:
- https://github.com/orgs/community/discussions/4324#discussioncomment-3477871

### DIFF
--- a/.github/workflows/pull-build-push.yaml
+++ b/.github/workflows/pull-build-push.yaml
@@ -20,33 +20,11 @@ on:
       - 'main'
 
 jobs:
-  filter-changes:
-    name: List relevant VCS changes
-    outputs:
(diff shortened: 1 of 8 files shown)


barchw (approved):
Gitleaks is not part of the success check, don't know if we care

Otherwise LGTM, the GH rules probably need to be updated for this to actually be mergable 🤦

PR #1422, opened 2024-11-04
Validation of path conflicts with `{**}` or `{*}` operators

barchw:
**Description**

I implemented the conflict validation based on the Trie data structure.

It does two things that are not done by a regular trie:
   - Nodes that are pointed to by `{**}` don't store their children, instead they store the path suffix that exists after `{**}`. Conflicts are checked by comparing the suffixes of paths with the same segments that precede `{**}`. This can be done, since the `{**}` operator must be the last in the path.
   - `{*}` nodes are stored just like any other exact segment node, but are always included in conflict search.
---

- Add conflict validation to envoy template paths

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/internal/path/segment_trie/segment_trie.go
+++ b/internal/path/segment_trie/segment_trie.go
@@ -55,7 +55,7 @@ func (t *SegmentTrie) InsertAndCheckCollisions(tokens []token.Token) error {
 
 	for i, tok := range tokens {
 		if tok.Type == token.BracedDoubleAsterix {
-			if _, ok := node.Children["{**}"]; !ok {
+			if n, ok := node.Children["{**}"]; !ok {
 				node.Children["{**}"] = &Node{
 					EndNode:  i == len(tokens)-1,
 					Children: make(map[string]*Node),
@@ -63,7 +63,7 @@ func (t *SegmentTrie) InsertAndCheckCollisions(tokens []token.Token) error {
 				}
 			} else {
 				suffixString := token.List(tokens[i:]).String()
-				node.Children["{**}"].Suffixes = append(node.Children["{**}"].Suffixes, suffixString)
+				n.Suffixes = append(n.Suffixes, suffixString)
 			}
 		} else {
 			if _, ok := node.Children[tok.Literal]; !ok {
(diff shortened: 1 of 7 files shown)


internal/path/segment_trie/segment_trie.go
```diff
@@ -0,0 +1,167 @@
...
+		return node.EndNode
+	}
+
+	if len(node.Suffixes) > 0 {
+		return hasAnySuffix(tokens, node.Suffixes)
+	}
+
+	switch tokens[cur].Type {
+	case token.Ident:
+		if _, ok := node.Children[tokens[cur].Literal]; ok {
```
Ressetkk, line 125: You reference the *value* of element you checked existence. Why don't you just use this value? This suggestion also applies to the IFs below.
```suggesion
		if val, ok := node.Children[tokens[cur].Literal]; ok {
					if findExistingPath(val, tokens, cur+1) {
				return true
			}}
```

I also don't like the nested value reference. I think for better readability you *should* assign current token you work with as `token := tokens[cur]` and then reference created variable in switch statement.

internal/path/segment_trie/segment_trie.go
```diff
@@ -0,0 +1,167 @@
...
+	if cnode, ok := node.Children["{**}"]; ok {
+		tokensString := token.List(suffix).String()
+		for _, suffix := range cnode.Suffixes {
+			if strings.HasSuffix(tokensString, suffix) || strings.HasSuffix(suffix, tokensString) {
+				return true
+			}
+		}
+	}
+
+	if _, ok := node.Children[suffix[cur].Literal]; ok {
```
Ressetkk, line 96: You reference the value of element you checked existence. Why don't you just use this value you blanked?

internal/path/segment_trie/segment_trie.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		}
+
+		if _, ok := node.Children["{**}"]; ok {
+			if findExistingPath(node.Children["{**}"], tokens, cur+1) {
+				return true
+			}
+		}
+	case token.BracedAsterix:
+		for _, node := range node.Children {
```
Ressetkk, line 143: this `node` assignment covers the function argument. For clearance, I'd change it to unique value.

Ressetkk (changes requested):

Ressetkk (approved):

Issue #1234, opened 2024-08-08
ci: store image artifacts in separate paths


Ressetkk:
/kind chore
/area ci

To unblock easy deployment of #1150 places where main and release images are put need to be separated.

Because we push images created by [post-api-gateway-manager-build](https://github.com/kyma-project/test-infra/blob/9598799dc949b1b5417d0e9075b04ee304081946/prow/jobs/kyma-project/api-gateway/api-gateway-manager-build.yaml#L59) and [rel-api-gateway-manager-build](https://github.com/kyma-project/test-infra/blob/9598799dc949b1b5417d0e9075b04ee304081946/prow/jobs/kyma-project/api-gateway/api-gateway-manager-build.yaml#L112C17-L112C42) to the same directory with the same build tag format `europe-docker.pkg.dev/kyma-project/prod/api-gateway-manager:vDATE-ShortSHA`. That results in mix-matching the images used in main and in production. It also complicates developing simple automations around our images if we wanted to keep implementations of automated tasks separated.
My suggestion is to separate main and release images with subpath:
* General idea: `europe-docker.pkg.dev/kyma-project/prod/{REPOSITORY}/{TYPE}/{NAME}`
* main images build to `europe-docker.pkg.dev/kyma-project/prod/api-gateway/main/api-gateway-manager`
* release images build to `europe-docker.pkg.dev/kyma-project/prod/api-gateway/releases/api-gateway-manager`

**This will not affect the way how our images are signed or validated, as validation is configured against `europe-docker.pkg.dev/kyma-project/prod`**

PR #1434, opened 2024-11-07
chore: separate main and release images

Ressetkk:
/kind chore
/area ci

Separate images that we run on main/releases/pr in order to keep better isolation between the artifacts to allow introducing registry automations.

#1234


triffer (commented):
I think we also need to change those images: https://github.com/kyma-project/api-gateway/blob/8ba8c22aac01bb4c7e498300da7186c04150a7c2/.github/workflows/ui-tests-periodic.yaml#L35

triffer (approved):

PR #1459, opened 2024-11-19
Use latest Busola version in UI tests

triffer:
**Description**

- Update Busola version used in UI tests to see if this helps with flakiness

**Related issues**


triffer:
Latest Busola version makes it flakiness even worse and causes a new test failure.

PR #1464, opened 2024-11-19
Fixed support for using oauth2-mock in custom domain tests

mluk-sap:
**Description**

- Fixed support for using oauth2-mock in custom domain tests
  - oauth2 mock is deployed after test namespace is created
  - removed scope from apirule configuration (so the scope is not required when requesting the token, which is more compatible with external OIDC)
  - adapted validation in env_vars_custom_domain script, to it is possible to run custom domain scripts via make (without OIDC)

**Pre-Merge Checklist**

**Related issues**


triffer (approved):
Thanks for improving those tests! They usually don't get much love, especially not when it comes to running them locally. 😄

PR #1489, opened 2024-11-27
feat: workflow to add dependabot PRs to project

Ressetkk:
/kind feature
/area ci

Run workflow on pull request to automate adding dependabot PRs to board.


.github/workflows/pr-add-dependabot-to-project.yaml
```diff
@@ -0,0 +1,104 @@
+on:
```
triffer, line 1: Just saw we have no name set on those jobs, so the filepath + file name will be used as name and that doesn't look very nice.

triffer (approved):

PR #1491, opened 2024-11-27
Test sanity checks

mluk-sap:
**Description**

- added a ValidateAndFixConfig function to the Testsuite interface, so each suite may now verify the (common) configuration using own rules and enforce some settings (like Gardener usage or Test concurrency=1)
- added validations:
  - check if test domain is set
  - check is custom domain is set in Custom Domain tests
  - check is GCP Service Account json path is set in in Custom Domain tests
  - check if client id and secret is set when using OIDC
  - check whether Gardener contains proper CRDs
  - check if test domain matches the Gardener domain in shoot-info
- validate if IMG is properly set in some Makefile tasks
- moved GCP Service Account secret creation to the tests (instead of Makefile), so it works properly when you run tests via IDE
- added TEST_DEBUG_LOGGING that enables some more verbose logs (like when an retryable action fails)
- moved OAuth server configuration into pkg, so it is reused in all suites
- moved namespace creation/deletion to pkg, so it is reused in all suites
- wait for oauth mock deployment to be fully running
- wait for ext-authz deployment to be fully running
- deploy oauth mock with unique host name, so there is a separation between test runs
- oauth mock is now a deployment, not a bare pod
- fail fast if api gateway exists
- report when operations are finished properly, so the logs are less confusing and it is clear what is really being executed at the moment
- empty config optional variables are now really empty instead of having "empty" value

**Pre-Merge Checklist**

**Related issues**


tests/integration/testsuites/custom-domain/manifests/resources.yaml
```diff
@@ -11,7 +22,6 @@ spec:
   type: google-clouddns
   secretRef:
     name: google-credentials
-    namespace: default
```
kolodziejczak, line 14: Should we drop the namespace field here? In the Secret defined above, there is still a {{.Namespace}} . Is the default for this field the namespace of the DNSProvider resource?

tests/integration/pkg/testcontext/config_validation.go
```diff
@@ -0,0 +1,98 @@
...
+	return nil
+}
+
+func (c *Config) ValidateGardener(resourceMgr *resource.Manager, k8sClient dynamic.Interface) error {
+	if c.IsGardener {
+		err := helpers.CheckGardenerSetup(resourceMgr, k8sClient)
+		if err != nil {
+			return fmt.Errorf("requested Gardener cluster, but the current cluster is not properly installed: %w", err)
+		}
+		if strings.Contains(c.Domain, localDomain) {
```
kolodziejczak, line 46: What if the `c.Domain` is set to something else than `local.kyma.dev`? For example `asd.example`. Is there a way to ensure in this part of the code, that the domain is Gardener's?

tests/integration/testsuites/custom-domain/manifests/resources.yaml
mluk-sap: Yes, so the secret must have a namespace field, so it is created in this namespace. Previously it was always created in the 'default' namespace.
In the DNSProvider we now reference the google-credentials secret *in the same namespace*, so we don't need to specify it. Previously the secret was in the different namespace ('default' one), so it was a cross-namespace reference, thus namespace had to be provided.

tests/integration/pkg/testcontext/config_validation.go
mluk-sap: We are considering only tests here. So there are only few cases possible:
- locally run tests, so everything works on localhost, that is why we have *.local.kyma.dev pointing to 127.0.0.1
- tests  run against the Gardener cluster using default cluster domain, which is stored in shoot info, so we validate whether they match
- some custom setup with custom domain - but this is done in custom-domain tests and then the additional domain is provided as a separate configuration attribute and not validated in any way, because it is completely different than the default Gardener one

kolodziejczak (approved):

tests/integration/pkg/testcontext/config_validation.go
mluk-sap: Okay, so we decided to change the logic here and auto detect the Gardener cluster, so the validation is useful also when an end-developer doesn't set IS_GARDENER=true

PR #1507, opened 2024-12-02
QOL updates to APIRule v2alpha1 integration tests

barchw:
**Description**

- Remove prefix initialiser, switching to universal initialisers
- Fix generation of namespace name to use a singular instance of RNG, fixing potential namespace conflicts
- Add a separate optional pre-start hook that scales down oathkeeper, masively speeding up APIGateway startup time
- Do not teardown when the test fail, allowing for post-run troubleshooting 

**Pre-Merge Checklist**

**Related issues**


tests/integration/pkg/hooks/api-gateway.go
```diff
@@ -114,7 +137,7 @@ var ApplyAndVerifyApiGatewayCrSuiteHook = func() error {
 		}
 
 		return nil
-	}, testcontext.GetRetryOpts()...)
+	}, retry.DelayType(retry.FixedDelay), retry.Delay(500), retry.Attempts(1200))
```
triffer, line 140: This seems like a very long time to retry the scale down. Why did we chose to make it that long?

tests/integration/pkg/hooks/api-gateway.go
```diff
@@ -69,7 +70,7 @@ var ApiGatewayCrTearDownScenarioHook = func(ctx context.Context, sc *godog.Scena
 	return ctx, nil
 }
 
-var ApplyAndVerifyApiGatewayCrSuiteHook = func() error {
+func ApplyAndVerifyApiGateway(scaleDownOathkeeper bool) error {
```
triffer, line 73: I think for a an easier to understand API it might be better to not export this function.

triffer (dismissed):
Very good changes, thanks for taking the time to clean this up. 👍

tests/integration/pkg/hooks/api-gateway.go
barchw: I will revert that to the old value

triffer (approved):

Issue #1516, opened 2024-12-04
RateLimit reconciliation


strekm:
**Description**

Implement reconcile function that uses EnvoyBuilder to manage RateLimit CR. Based on configuration in RateLimit corresponding Istio EnvoyFilter should be created. RateLimit Status should be set accordingly. APIGateway blocking deletion mechanism should be extended to block deletion if any RateLimit resources are created on a cluster. e2e test is out of scope in this task

**ToDos [Developer]**

- [x] Check if https://github.com/Ressetkk/api-gateway/commit/0b8400cd7a7ef8301bcc4f618fa1b65a16bbf189 could be a viable option for reconcile loop (try to implement it according to this proposal, meet and present) - 1d timebox, drop if cannot be reasonably used
- [x]  Apply EnvoyFilter in the cluster according to configuration based on RateLimit
- [x]  Status handling
- [ ] ~~Extend API-gateway blocking deletion~~ Will be handled in a followup

**PRs**

- [x] https://github.com/kyma-project/api-gateway/pull/1543
- [x] https://github.com/kyma-project/api-gateway/pull/1558

**ACs [PO]**

- [x] RateLimit CR being reconciled
- [x] corresponding Istio EnvoyFilter is created
- [x] Status is set
- [x] whole functionality is hidden behind feature toggle
- [ ] ~~APIGateway deletion is blocked if any RateLimit CRs are created~~

**DoD [Developer & Reviewer]**

- [x] Provide unit and integration tests.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed.
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.

**Attachments**

- part of: https://github.com/kyma-project/api-gateway/issues/1365

PR #1558, opened 2024-12-23
bug: RateLimit: replace Watch with polling

Ressetkk:
/kind bug
/area api-gateway

* Requeue reconciliation every 3 minutes, accepting downtime
* Add warning state to the CR status
* Adjust tests

Restore watching should be handled in a follow-up anytime soon.

#1516

### DIFF
--- a/apis/gateway/ratelimit/v1alpha1/ratelimit_types.go
+++ b/apis/gateway/ratelimit/v1alpha1/ratelimit_types.go
@@ -21,8 +21,9 @@ import (
 )
 
 const (
-	StatusReady = "Ready"
-	StatusError = "Error"
+	StatusReady   = "Ready"
+	StatusWarning = "Warning"
+	StatusError   = "Error"
 )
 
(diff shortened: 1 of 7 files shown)


videlov (dismissed):

main.go
```diff
@@ -245,7 +245,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = ratelimit.Setup(mgr, scheme); err != nil {
+	if err = ratelimit.Setup(mgr, 3*time.Minute, scheme); err != nil {
```
barchw, line 248: Please make that a configurable parameter / flag / const

barchw (changes requested):

barchw (approved):
Screenshot 2024-12-23 at 15 56 43
Looks much better, TY
Don't know if it should be an `Error` though? From the Kyma Module api standpoint it's more in the `warning`, user action territory

mluk-sap (approved):

PR #1569, opened 2024-12-27
Full support for special characters in asterisk paths

barchw:
**Description**

- Full support for special characters in asterisk paths
- Fix support for space in url (`%20` and `+`)

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/internal/processing/processors/v2alpha1/virtualservice/virtual_service_processor.go
+++ b/internal/processing/processors/v2alpha1/virtualservice/virtual_service_processor.go
@@ -20,8 +20,8 @@ const defaultHttpTimeout uint32 = 180
 
 var (
 	envoyTemplatesTranslation = map[string]string{
-		`{**}`: `[\w\.~\-\/]*`,
-		`{*}`:  `[\w\.~\-]*`,
+		`{**}`: `([A-Za-z0-9-._~!$&'()*+,;=:@/]|%[0-9a-fA-F]{2})*`,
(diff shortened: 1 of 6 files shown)


internal/processing/processors/v2alpha1/virtualservice/virtual_service_processor.go
```diff
@@ -20,8 +20,8 @@ const defaultHttpTimeout uint32 = 180
 
 var (
 	envoyTemplatesTranslation = map[string]string{
-		`{**}`: `[\w\.~\-\/]*`,
-		`{*}`:  `[\w\.~\-]*`,
+		`{**}`: `([A-Za-z0-9-._~!$&'()*+,;=:@/]|%[0-9a-fA-F]{2})*`,
+		`{*}`:  `([A-Za-z0-9-._~!$&'()*+,;=:@]|%[0-9a-fA-F]{2})+`,
```
Ressetkk, line 24: Arę you sure we don't want to use the built-in function that encodes URI? https://pkg.go.dev/net/url#PathEscape. I would be very cautious to write it on our own. Otherwise no additional comments.

barchw: Unfortunately, I think that there is no other option as this thing here is just passed on to the VirtualService. We don't really do the validation here, only provide the way to do it

Ressetkk (approved):

PR #1581, opened 2025-01-03
Bug fix: don't constantly update APIRule in tests while getting the status

mluk-sap:
**Description**

- Refactored the way of applying apiRules in tests, so the behavior is more explicit and predictable:
  - introduced a function to read single manifest template from yaml file (returns error if there are multiple manifests)
  - simplified APIRule helper functions, so they do what is expected:
    - applyApiRule creates new single ApiRule (returns error if APIRule exists)
    - updateApiRule updates an existing single ApiRule (returns error if APIRule doesn't exist)
    - these functions don't call UpdateResources anymore in order to get APIRule status, which was breaking tests a lot
  - istio-jwt / scenario_noauth_and_jwt - proper installation of two APIRules (previously only one was considered in state analysis)
  - ory / service_per_path - helloworld workload deployed explicitly (previously it was deployed as a side effect of applying APIRule)
- Reintroduced ScenarioInitializer for all tests, which improves test separation (regenerates random identifiers) and allows to detect cases where tests share some state in an implicit way (previously Custom Domain tests were sharing the same APIRule name)
- Refactored Custom Domain tests (because of ScenarioInitializer changes):
  - custom domain resources are prepared in BeforeSuiteHooks and cleaned in AfterSuiteHooks
  - scenarios are now separated, so they don't share any state, which is a good practice (previously APIRule was common)
  - improved readability of logs - more logs during preparation, test output is not convoluted by log messages, etc.
  - renamed some variables, so their purpose is more clear (like subdomainForTests)
- Improved APIRule related error messages, so they contain more context information (like APIRule name)

**Pre-Merge Checklist**

**Related issues**


Ressetkk (approved):
lgtm. Thanks.

tests/integration/pkg/helpers/api_rule.go
```diff
@@ -63,47 +61,55 @@ func getAPIRuleStatus(res *unstructured.Unstructured) (string, string, error) {
...
 	return code, description, nil
 }
 
-// ApplyApiRule tries toExecute function and retries with onRetry if APIRule status is in error status or has no status
-// code. This function works for both v1beta1 and v2alpha1 versions of APIRule.
-func ApplyApiRule(toExecute RetryableApiRule, onRetry RetryableApiRule, k8sClient dynamic.Interface, retryOpts []retry.Option, resources []unstructured.Unstructured) error {
-	res, err := toExecute(k8sClient, resources...)
+// ApplyApiRule creates APIRule and waits for its status
+// This function works for both v1beta1 and v2alpha1 versions of APIRule.
+func ApplyApiRule(resourceMgr *resource.Manager, k8sClient dynamic.Interface, retryOpts []retry.Option, apiRuleResource unstructured.Unstructured) error {
```
barchw, line 72: We can switch to `CreateApiRule` as discussed offline

tests/integration/pkg/network/dns.go
```diff
@@ -66,3 +70,31 @@ func restartCoreDnsPods() error {
 
 	return c.Patch(context.Background(), dep, patch)
 }
+
+// function waits until the given domain points to the given IP
+// the domain must be a wildcard one because the function probes the <randomHost>.<domain>
+func WaitUntilDNSReady(domain string, expectedIP net.IP) error {
+	err := wait.ExponentialBackoff(wait.Backoff{
```
barchw, line 77: Looks like this might be taking to much time, I would consider fixed delay here

barchw (changes requested):

barchw (approved):

kolodziejczak (approved):

Issue #1626, opened 2025-01-20
bug: RateLimit: fail when fill_interval is set above `60s`


Ressetkk:
/kind bug
/area api-gateway

**Description**

When RateLimit is deployed, Istiod returns an error if `fill_interval` contains values below or above seconds.
```
2025-01-20T10:50:32.427568Z     warn    delta   ADS:LDS: ACK ERROR httpbin.test-1 Internal:Error adding/updating listener(s) virtualInbound: Unable to parse JSON as proto (INVALID_ARGUMENT: invalid JSON in   envoy.extensions.filters.http.local_ratelimit.v3.LocalRateLimit @ descriptors[0].token_bucket.fill_interval: message google.protobuf.Duration, near   1:514 (offset 513): duration must end   with  a single 's'): {...}
```

**Expected result**
RateLimit configuration gets applied correctly.

**Steps to reproduce**

RateLimit CR
```yaml
apiVersion: gateway.kyma-project.io/v1alpha1
kind: RateLimit
metadata:
  labels:
    app: httpbin
  name: ratelimit-path-sample
  namespace: test
spec:
  selectorLabels:
    app: httpbin
  enableResponseHeaders: true
  local:
    defaultBucket:
      maxTokens: 100
      tokensPerFill: 50
      fillInterval: 30s
    buckets:
      - path: /ip
        bucket:
          maxTokens: 1
          tokensPerFill: 1
          fillInterval: 60m
```

**Troubleshooting**
I believe we can't rely on go's default duration parse. All durations need to be converted to seconds. When EnvoyFilter contains `ms`, I got an error:
```
2025-01-20T13:13:42.256815Z     warn    delta   ADS:LDS: ACK ERROR httpbin.test-17 Internal:Error adding/updating listener(s) virtualInbound: Proto constraint validation failed (LocalRateLimitValidationError.Descriptors[0]: embedded message failed validation | caused by LocalRateLimitDescriptorValidationError.TokenBucket: embedded message failed validation | caused by TokenBucketValidationError.FillInterval: value must be greater than 0s): stat_prefix: "rate_limit"
```

That means we can only use seconds.

mluk-sap:
PR: https://github.com/kyma-project/api-gateway/pull/1627

PR #1627, opened 2025-01-20
bug: fix duration parse in ratelimit.go

Ressetkk:
/kind bug
/area api-gateway

Fix duration parsing in RateLimit's EnvoyFilter patch
* Project value as seconds in EnvoyFilter
* accept only interval greater or equal than 50ms
* Add validation to check if fill_intervals for all buckets is greater than 0s

#1626


videlov:
in the ADR `fillInterval` is defined as 50ms or more:
https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0011-rate-limit.md

should we update that as well since we are changing it in the code?

internal/ratelimit/validate_test.go
```diff
@@ -303,4 +304,49 @@ var _ = Describe("RateLimit CR Validation", func() {
...
+					"app": "test",
+				},
+				Local: ratelimitv1alpha1.LocalConfig{DefaultBucket: ratelimitv1alpha1.BucketSpec{FillInterval: &metav1.Duration{Duration: 0}}},
+			},
+		}
+		c := fake.NewClientBuilder().WithScheme(sc).WithObjects(&rlCR).Build()
+
+		err := ratelimitvalidator.Validate(context.Background(), c, rlCR)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("fill_interval must be greater or equal 50ms"))
```
videlov, line 324:
```suggestion
		Expect(err.Error()).To(ContainSubstring("default_bucket: fill_interval must be greater or equal 50ms"))
```

videlov (approved):

PR #1630, opened 2025-01-22
Proposal of handling of module status

barchw:
**Description**

- Proposal of handling of module status

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/docs/contributor/adr/0011-module-status.md
+++ b/docs/contributor/adr/0011-module-status.md
@@ -1,0 +1,41 @@
+# Module Status Handling
+
+## Status
+Proposed
+
+## Context
+
+Currently, APIGateway module sets the state `Processing` on the APIGateway Custom Resource whenever there is reconciliation / installation happening.
+From the monitoring and kubernetes API standpoint, this is not easy to properly handle,
+as there is no possibility to observe module readiness, signified by module CR being in `Ready` state,
+without having in mind the periodic switch to the `Processing` state.
+
+## Proposal
+
+This ADR proposes changes to handling the state of the module CR, with three different solutions possible:
+
+### Solution 1
+
+Proposal: Remove the `Processing` state entirely from the module API.
+
+Consequences: From technical standpoint, this would be a breaking change, as there might be users relying on the `Processing` state,
+for example, to determine if the module is being installed or not.
+However, as the purpose of the `Processing` state is not clear, it might be a good idea to remove it entirely.
+
+### Solution 2
+
+Proposal: Do a `soft` removal of the `Processing` state. 
+This would mean that this state is still present in the API, but will never be set by the module.
+
+Consequences: This would be a less breaking change than the first solution, as the `Processing` state would still be present in the API,
+but the installation would not be possible to observe in this state.
+
+### Solution 3
+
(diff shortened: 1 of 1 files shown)


Ressetkk:
I think ADR should be focused around one solution that is already the one that we want to implement, rather than presenting possible solutions. That creates open questions which are in contradiction with Architecture Decision Record, because the architecture decision is yet to be made. I think we should analyse each of the solutions, present results, decide which solution to accept, and then prepare a decision record.

Moreover, I think we should really think again how our operator handles resources. Maybe it'd be better to just reduce the dependency in APIGateway CR controller to actually behave as operator which manages the lifecycle of controllers, and maybe move some of the logic to separate controller...

Moving back to the topic - usually the status keeps the finite state of a resource and changes if the said state changes. That means setting a status rather should be a final step in resource reconciliation. Keeping `Processing` as transient state will make a problem with resource conflicts, if you for example set this state at the beginning of reconciliation, then you have to `Get()` resource again to fetch its updated state.

Ideally this should look like that:
1. User applies a CR
2. Controller schedules bunch of stuff
3. Controller sets up `Processing` status, finishes reconciliation and queues the next run in seconds
4. In the next run controller checks the state of each of the stuff we scheduled
5. If they are working/done, controller sets  state to `Ready` and finishes work
6. If they failed, controller sets state to `Error/Warning` and finishes work - user attention is needed
7. If they still run, controller leaves `Processing` status and queues next run.

strekm (dismissed):

mluk-sap:
Let's rework a document a little bit, so the proposed solution is described in details in the main section (like 'Decision') and put other ones in 'considered alternative solutions'.

docs/contributor/adr/0011-module-status.md
```diff
@@ -0,0 +1,52 @@
...
+## Context
+
+Currently, APIGateway module sets the state `Processing` on the APIGateway Custom Resource whenever there is reconciliation / installation happening.
+From the monitoring and kubernetes API standpoint, this is not easy to properly handle,
+as there is no possibility to observe module readiness, signified by module CR being in `Ready` state,
+without having in mind the periodic switch to the `Processing` state.
+
+## Decision
+
+This ADR proposes changes to handling the state of the module CR, with three different solutions possible:
```
mluk-sap, line 15: We need to adapt this to the new structure with alternative solutions.

docs/contributor/adr/0011-module-status.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Solution 2
+
+Proposal: Do a `soft` removal of the `Processing` state. 
+This would mean that this state is still present in the API, but will never be set by the module.
+
+Consequences: This would be a less breaking change than the first solution, as the `Processing` state would still be present in the API,
+but the installation would not be possible to observe in this state.
+
+### Solution 3
```
mluk-sap, line 38: This should be moved to 'Decision', because it is not alternative one, but the chosen one, right?

nataliasitko (approved):

mluk-sap (approved):

pbochynski (approved):
LGTM

Issue #1629, opened 2025-01-21
bug: ratelimit: validation error `path or headers must be set` when both path and headers fields are present in CR


Ressetkk:
When the following rate limit CR is applied the the validation returns an error:
```
The RateLimit "ratelimit-header-sample" is invalid: spec.local.buckets[0]: Invalid value: "object": path or headers must be set
```

```yaml
apiVersion: gateway.kyma-project.io/v1alpha1
kind: RateLimit
metadata:
  labels:
    app: httpbin
  name: ratelimit-header-sample
  namespace: test
spec:
  selectorLabels:
    app: httpbin
  enableResponseHeaders: true
  local:
    defaultBucket:
      maxTokens: 1
      tokensPerFill: 1
      fillInterval: 30s
    buckets:
      - headers:
          X-Rate-Limited: "true"
        path: /headers
        bucket:
          maxTokens: 10
          tokensPerFill: 5
          fillInterval: 30s
```

This is incorrect behaviour, because this particular bucket example is projected as valid in ADR. See: https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0011-rate-limit.md#usage-example

The error is in validation rule in https://github.com/kyma-project/api-gateway/blob/a936d168d69da0ab87761c35b58c1a3769015cac/apis/gateway/ratelimit/v1alpha1/ratelimit_types.go#L30. It only allows either `path` or `headers` to be defined.

/kind bug
/area api-gateway

mluk-sap:
Done
PR: https://github.com/kyma-project/api-gateway/pull/1633

PR #1633, opened 2025-01-22
bug: fix bucket validation in ratelimit_types.go

Ressetkk:
As defined in [ADR](https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0011-rate-limit.md#usage-example), allow scenario where both `path` and `headers` can be defined for single token bucket.
Cover with controller integration tests.

> It is also possible to combine request descriptors, for example rate limit on path and headers, in the same rate limit configuration. In this case the rate limit is applied only if all specified descriptors are present in the request.

/kind bug
/area api-gateway

#1629


kubicar (approved):
LGTM 💪

Issue #1609, opened 2025-01-15
rate limit e2e test


strekm:
**Description**

Implement automated e2e test testing limiting access to a workflow, based on scenario described in https://github.com/kyma-project/api-gateway/issues/1607.  Focus on testing following scenarios:
- connectivity to test pod is rate-limited when default bucket is empty (maxTokens: 0)
- connectivity to test pod on custom endpoint is rate-limited when path bucket is empty
- path based rate limit
- request header based rate limit

Test should deploy httpbin and expose using APIRule v2alpha1. Access to the httpbin should be limited by creating corresponding RateLimit CR. 

**PRs**

- [x] #1641 

**ACs [PO]**

- [x] separate test suite for rate limit e2e scenario
- [x] test executed on k3d and gardener clusters
- [x] testing path based rate limit
- [x] testing request header rate limit
- [x] if needed documentation extended (not required)

**DoD [Developer & Reviewer]**

- [x] Provide unit and integration tests.
- [x] Provide documentation. (not required)
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed. (not applicable)
- [x] If the default configuration of Istio Operator has been changed, you performed a manual upgrade test to verify that the change can be rolled out correctly. (not applicable)
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [x] Add release notes. (not applicable)

**Attachments**

- [...]()

**Follow-Up Issues**

- part of #1365

mluk-sap:
Done
PR: https://github.com/kyma-project/api-gateway/pull/1641

Issue #1607, opened 2025-01-15
Rate limit tutorial


strekm:
**Description**

Create tutorials explaining how to expose and limit access to a workload. Tutorials should match style of other tutorials. One tutorial should focus on how to configure RateLimit CR to limit access to a workload based on path. Second tutorial should highlight on limiting access based on given request header. Each of tutorial should contain a note referring to https://github.com/kyma-project/api-gateway/issues/1366. 

Open question:
- which version of APIRule should be used -> decision use v2alpha1

**ToDos [Developer]**

- [x] Create a tutorial:
 - Create test Istio-injected namespace
 - Deploy httpbin pod + service
 - Expose httpbin via APIRule (basic, no_auth)
 - Create RateLimit CR with local configuration for two scenarios and enabled response headers:
   * Default bucket (all paths)
   * Exact path (/ip)
   * Request headers (x-test-header)
   * Path AND request headers configured (/ip and x-test-header)
   * All scenarios *must* have different RateLimit configurations
 - Use curl to confirm that the configuration is working

**PRs**

- [x] https://github.com/kyma-project/api-gateway/pull/1631

**ACs [PO]**

- [x] path based rate limit tutorial added
- [x] request header rate limit tutorial added
- [x] both tutorials contain noted described in https://github.com/kyma-project/api-gateway/issues/1366

**DoD [Developer & Reviewer]**

- [x] Provide documentation.

**Attachments**

- part of #1365

Issue #1608, opened 2025-01-15
RateLimit feature toggle removal


strekm:
**Description**

Remove feature toggle guarding rate limiting feature.

**ToDos [Developer]**

~- [ ] remove build task~ 
- [x] remove ratelimit_disabled file 
- [x] adjust make / docker file if need
- [x] remove ratelimit const
- [x] remove ratelimit const from tests

**PRs**

- [x] https://github.com/kyma-project/api-gateway/pull/1644

**ACs [PO]**

- [x] rate limit functionality exposed
- [x] release notes prepared

**DoD [Developer & Reviewer]**

- [x] Provide unit and integration tests.
- [x] Provide documentation.
- [x] Verify if the solution works for both open-source Kyma and SAP BTP, Kyma runtime.
- [x] If you changed the resource limits, explain why it was needed.
- [x] If the default configuration of Istio Operator has been changed, you performed a manual upgrade test to verify that the change can be rolled out correctly.
- [x] Verify that your contributions don't decrease code coverage. If they do, explain why this is the case.
- [x] Add release notes.

**Attachments**

- part of #1365

kolodziejczak:
Done with with https://github.com/kyma-project/api-gateway/pull/1644

PR #1641, opened 2025-01-27
test: rate limit e2e tests

kubicar:
**Description**

- added e2e tests for rate limit CR 

**Pre-Merge Checklist**

**Related issues**
#1609  #1607

### DIFF
--- a/tests/integration/pkg/helpers/http_client.go
+++ b/tests/integration/pkg/helpers/http_client.go
@@ -117,10 +117,6 @@ func (h *RetryableHttpClient) withRetries(httpCall func() (*http.Response, error
 	return nil
 }
 
-func (h *RetryableHttpClient) GetHttpClient() *http.Client {
-	return h.client
-}
-
 type OIDCConfiguration struct {
 	Issuer                                string   `json:"issuer"`
 	AuthorizationEndpoint                 string   `json:"authorization_endpoint"`
--- a/tests/integration/pkg/helpers/http_client.go
+++ b/tests/integration/pkg/helpers/http_client.go
@@ -117,6 +117,10 @@ func (h *RetryableHttpClient) withRetries(httpCall func() (*http.Response, error
 	return nil
 }
 
+func (h *RetryableHttpClient) HttpClient() *http.Client {
+	return h.client
+}
+
 type OIDCConfiguration struct {
 	Issuer                                string   `json:"issuer"`
 	AuthorizationEndpoint                 string   `json:"authorization_endpoint"`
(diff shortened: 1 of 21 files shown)


kubicar:
waiting for #1608

tests/integration/pkg/helpers/http_client.go
```diff
@@ -117,6 +117,10 @@ func (h *RetryableHttpClient) withRetries(httpCall func() (*http.Response, error
 	return nil
 }
 
+func (h *RetryableHttpClient) GetHttpClient() *http.Client {
```
Ressetkk, line 120: It's unnecessary for getters to start with Get.

https://go.dev/doc/effective_go#Getters

tests/integration/testsuites/rate-limit/testsuite.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/kyma-project/api-gateway/tests/integration/pkg/global"
+	"github.com/kyma-project/api-gateway/tests/integration/pkg/helpers"
+	"github.com/kyma-project/api-gateway/tests/integration/pkg/resource"
+	"github.com/kyma-project/api-gateway/tests/integration/pkg/testcontext"
+)
+
+const manifestsDirectory = "testsuites/rate-limit/manifests/"
+
+func (t *testsuite) createScenario() *scenario {
+	testId := helpers.GenerateRandomTestId()
```
Ressetkk, line 20: When you declare variable with shorthand in it the shorthand should be uppercase.
```suggestion
	testID := helpers.GenerateRandomTestID()
```

tests/integration/testsuites/rate-limit/scenario.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	httpClient := s.httpClient.GetHttpClient()
+	req := &http.Request{
+		URL:    endpointUrl,
+		Method: method,
+		Header: map[string][]string{
+			"X-Rate-Limited": {"true"},
+		},
+	}
+	for i := 0; i < n; i++ {
```
Ressetkk, line 44: I wonder if instead of hard two hard request calls, we should implement the eventuality and exponential back off with a timeout using `request.Do()`. I'm afraid it might result in flakes when connectivity in the cluster is unstable.

Ressetkk (changes requested):

tests/integration/testsuites/rate-limit/scenario.go
kubicar: I that case (connectivity in the cluster is unstable) yes it can help, but when you will get connection error and apply exponential back off then counting of requests is useless 🤔. But it should help with scenario when we experience this connectivity issue on the beginning when pod is not ready.

PR #1657, opened 2025-02-03
add-release-notes

nataliasitko:
**Description**

- ...

**Pre-Merge Checklist**

**Related issues**


docs/release-notes/2.10.3.md
```diff
@@ -0,0 +1,5 @@
+## New Features
+If you expose a path using a wildcard pattern, now you can also expose the same path without including the trailing slash. For example, if you have a rule that exposes `example.com/path/*`, you can also expose `example.com/path` without the slash at the end. ([#1625](https://github.com/kyma-project/api-gateway/pull/1625))
```
nataliasitko, line 2:
```suggestion
If you have exposed a path using a wildcard pattern, now you can also expose the same path without including the trailing slash. For example, if you have a rule that exposes `example.com/path/*`, you can also expose `example.com/path` without the slash at the end. ([#1625](https://github.com/kyma-project/api-gateway/pull/1625))
```

Ressetkk (approved):

PR #1716, opened 2025-02-18
feat: add Dockerfile.debug

Ressetkk:
This Dockerfile contains configuration for remote debugging session for the module.

It's only meant for running locally.
To use the feature
1. Build using `docker build`
2. Load to the cluster using `kind load image` or `k3d image import`
3. Change the image in the `containers` section to yor debug tag
4. Remove commands field from manager deployment
5. kubectl port-forward -n kyma-system {manager pod} 40000:40000 8000:8000

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -1,0 +1,42 @@
+# Build the manager binary
+FROM golang:1.23.5-alpine AS builder
+ARG GO_BUILD_TAGS
+ARG VERSION
+
+WORKDIR /api-gateway-build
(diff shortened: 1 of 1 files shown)


Dockerfile.debug
```diff
@@ -0,0 +1,42 @@
+# Build the manager binary
+FROM golang:1.23.5-alpine AS builder
```
kolodziejczak, line 2: can we go already with go 1.24?

kolodziejczak (changes requested):

Issue #1765, opened 2025-03-06
bug: apirule: controller does not update sub resources when workload service gets updated


Ressetkk:
**Description**

When underlying K8s service that APIRule exposes gets updated with new metadata, like labels or annotations, the APIRule does not update its subresources, for example AuthorizationPolicy does not get updated with new labels.

**Expected result**

APIRules update their sub resources when Kubernetes service gets updated.

**Actual result**
APIRule sub resources do not get updated.

**Steps to reproduce**
1. Deploy service that points to a workload
2. Expose that service with APIRule
3. Check if AuthorizationPolicy is created
4. Change the selector labels of Kubernetes service
5. Check if AuthorizationPolicy has updated selector labels

**Troubleshooting**

When APIRule gets updated, controller restarts, or APIRule is schedules for reconciliation, the problem is mitigated.

Possible solution: Set up a watch on Kubernetes service in APIRule controller. Then if the service triggers the reconciliation, list all APIRules in namespace (all namespaces?) and schedule reconcile for each of the APIRules, maybe updating a resource status to "Processing" or something else.

ingentismg:
Hi @Ressetkk,

not sure if this is a follow up to an SAP issue we created or just coincidence, but we are having the same issue.

I believe that there already is a reconciliation happening, although it is not very fast or happening very often. I did notice that after some time (that could be 10 minutes, 30 minutes, more; I am not sure as there is no status field on the resource) but at least 1 hour later the AuthroizationPolicy got updated.
For us, that is way too late. And for our production cluster we will have >800 AuthorizationPolicies, so this may be even slower.

One aspect of this I am currently feeling a bit uneasy about are APIRules that should protect a service using one of the auth feature, e.g. JWT.
That service would be exposed until the reconciliation happens.

Ressetkk:
> not sure if this is a follow up to an SAP issue we created or just coincidence, but we are having the same issue.

@ingentismg this issue is exactly a follow-up of your SNOW issue :D

> I believe that there already is a reconciliation happening, although it is not very fast or happening very often. I did notice that after some time (that could be 10 minutes, 30 minutes, more; I am not sure as there is no status field on the resource) but at least 1 hour later the AuthroizationPolicy got updated.

Yes, the periodic reconciliation is happening. Default requeue for APIRules is 30 minutes. As a workaround you can just restart api-gateway-controller-manager to requeue all reconciliations, or annotate APIRule.

[...]

PR #1773, opened 2025-03-10
bug: enqueue apirules reconcile on svc changes

Ressetkk:
* implement watch apirule controller to watch on service changes
* add test that checks if resource got updated

#1765


kubicar (approved):
checked and tested 👌 
Great job 💪

Issue #1800, opened 2025-03-26
AuthorizationPolicy controlled by APIRule should take host into consideration and should be applied only for the external traffic


strekm:
Current behavior of APIRule v2 is inconsistent when it comes to:
- handling host names - host names are set only in VirtualService, they are not used in AuthorizationPolicies, this is misleading
- detecting internal/external traffic - AuthorizationPolicy with noAuth is limited to traffic coming from Ingress, while the AuthorizationPolicy with jwt doesn't limit it - this is misleading, because we [documented that the internal traffic is blocked](https://github.com/kyma-project/api-gateway/blob/main/docs/user/custom-resources/apirule/04-70-changes-in-apirule-v2.md#internal-traffic-to-workloads-is-blocked-by-default) 

Handling host names:
Let's assume two APIRules:
- APIRule 1: host: foo-noauth.com, service: foo, path: *, method GET, access: noAuth
- APIRule 2: host: foo-jwt.com service: foo, path: *, method GET, access: jwt

Current behavior:
- foo-noauth.com - works without token
- foo-jwt.com - without token -> works - **this is misleading**
- foo-jwt.com - works with token
- foo cluster local traffic via internal host - is blocked (as intended)
- foo cluster local traffic via IP - is blocked (as intended)

Expected behavior:
- foo-noauth.com - should work (no change)
- foo-jwt.com - without token - should be blocked
- foo-jwt.com - with token - should work (no change)
- foo cluster local traffic via internal host - should be blocked (no change)
- foo cluster local traffic via IP - should be blocked (as intended)

Technically we should just copy ApiRule.spec.host to the AuthorizationPolicy.spec.rules.to.operation.host

Handling internal traffic:

Current behavior:
- APIRule with noAuth creates AP with source with principals entry istio-ingressgateway-service-account
- APIRule with jwt creates AP with source with requestPrincipals entry pointing to the provided jwt - which is not limited to ingress service account, so it opens it also for cluster local traffic - **this is misleading**

Expected behavior:
- APIRule should create AP that is always limited to the traffic coming from the Ingress

Note that the Istio has a [very specific way of applying rules](https://istio.io/latest/docs/reference/config/security/authorization-policy/):
- if there aren't any allow policies -> traffic is allowed by default
- if there is any allow policy -> traffic is denied by default
We documented this behavior recently, so the end-users should explicitly allow the internal traffic if they use APIRules.

TODO:
- [x] always limit AP to incoming traffic (source Ingress)
[...]

PR #1823, opened 2025-04-03
AuthorizationPolicy adjustments

Ressetkk:
- add host handling to authorizationPolicy
- adjust unit tests with adding a host
- copy over the code from virtualservice processor with a comment
- adjust in-cluster connectivity test to pass if in-cluster connectivity is blocked

Things that have not been changed:
- the sources for ory passthorugh APs since it's a special case during migration
- ExtAuth APs are not covered, because it requires additional work needed around lifecycle of the subresources

#1800

### DIFF
--- a/docs/release-notes/3.1.0.md
+++ b/docs/release-notes/3.1.0.md
@@ -1,3 +1,3 @@
-# Bug fixes
+# Bug Fixes
 
-- Fix behaviour of in-cluster connectivity when using JWT handler in APIRule. Now in-cluster connectivity is blocked. [#1823](https://github.com/kyma-project/api-gateway/pull/1823)
+- Fix behaviour of in-cluster connectivity when using JWT handler in APIRule. Now, in-cluster connectivity is blocked. [#1823](https://github.com/kyma-project/api-gateway/pull/1823)
--- a/docs/release-notes/3.1.0.md
+++ b/docs/release-notes/3.1.0.md
@@ -1,0 +1,3 @@
+# Bug fixes
+
+- Fix behaviour of in-cluster connectivity when using JWT handler in APIRule. Now in-cluster connectivity is blocked. [#1823](https://github.com/kyma-project/api-gateway/pull/1823)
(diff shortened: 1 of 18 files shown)


kubicar:
according to TODOs in issue #1800 ADR & user facing documentation is missing

joanbod:
Tested expected behavior on Gardener cluster using IAS as issuer for JWT:

- foo-noauth.com - should work (no change)
- foo-jwt.com - without token - should be blocked
- foo-jwt.com - with token - should work (no change)
- foo cluster local traffic via internal host - should be blocked (no change)
- foo cluster local traffic via IP - should be blocked (as intended)

Every case works as defined above.

kubicar (approved):

Ressetkk:
> according to TODOs in issue #1800 ADR & user facing documentation is missing

There is a [document](https://github.com/kyma-project/api-gateway/blob/main/docs/user/custom-resources/apirule/04-70-changes-in-apirule-v2.md) describing changes in apirule v2 that include internal traffic behaviour. There is also a [migration guide](https://github.com/kyma-project/api-gateway/blob/main/docs/user/custom-resources/apirule/04-60-apirule-migration.md) mentioning that in-cluster connectivity is blocked. There will also be a troubleshooting guide that should also include the information and solution on that.

kubicar (dismissed):

docs/release-notes/3.1.0.md
```diff
@@ -0,0 +1,3 @@
+# Bug fixes
+
+- Fix behaviour of in-cluster connectivity when using JWT handler in APIRule. Now in-cluster connectivity is blocked. [#1823](https://github.com/kyma-project/api-gateway/pull/1823)
```
nataliasitko, line 3:
```suggestion
- Fix behaviour of in-cluster connectivity when using JWT handler in APIRule. Now, in-cluster connectivity is blocked. [#1823](https://github.com/kyma-project/api-gateway/pull/1823)
```

docs/release-notes/3.1.0.md
```diff
@@ -0,0 +1,3 @@
+# Bug fixes
```
nataliasitko, line 1:
```suggestion
# Bug Fixes
```

nataliasitko (approved):

kubicar (approved):

PR #1837, opened 2025-04-08
fix: add deprecation warning for APIRule v1beta1 version

kubicar:
**Description**

- includes changes to mark the `APIRule` version as deprecated and provide warnings to users. 

**Pre-Merge Checklist**

**Related issues**

### DIFF
--- a/apis/gateway/v1beta1/apirule_types.go
+++ b/apis/gateway/v1beta1/apirule_types.go
@@ -85,6 +85,7 @@ func (s *APIRuleStatus) ApiRuleStatusVersion() versions.Version {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.APIRuleStatus.code"
 // +kubebuilder:printcolumn:name="Host",type="string",JSONPath=".spec.host"
+// +kubebuilder:deprecatedversion:warning="This version of APIRule is deprecated and will be removed in future releases. Please use v2 instead."
 type APIRule struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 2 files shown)


apis/gateway/v1beta1/apirule_types.go
```diff
@@ -85,6 +85,7 @@ func (s *APIRuleStatus) ApiRuleStatusVersion() versions.Version {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.APIRuleStatus.code"
 // +kubebuilder:printcolumn:name="Host",type="string",JSONPath=".spec.host"
+// +kubebuilder:deprecatedversion:warning="This version of APIRule is deprecated and will be removed in future releases. Please use v2 instead."
```
barchw, line 88: Those deprecation notes should include which version they are pointing to, as generally it might not be visible by the user during apply, and instead e.g. when checking for general api deprecations on cluster
```suggestion
// +kubebuilder:deprecatedversion:warning="v1beta1 version of APIRule is deprecated and will be removed in future releases. Please use v2 instead."
```

config/crd/bases/gateway.kyma-project.io_apirules.yaml
```diff
@@ -23,6 +23,9 @@ spec:
     - jsonPath: .spec.host
       name: Host
       type: string
+    deprecated: true
+    deprecationWarning: This version of APIRule is deprecated and will be removed
+      in future releases. Please use v2 instead.
```
nataliasitko, line 28:
```suggestion
    deprecationWarning: Version v1beta1 of APIRule is deprecated and will be removed
      in future releases. Use version v2 instead.
```

barchw (approved):

nataliasitko (approved):

PR #1869, opened 2025-04-22
bug: zero downtime migration fixup

Ressetkk:
do not add rules to the rule during migration step


internal/processing/processors/v2alpha1/authorizationpolicy/creator.go
```diff
@@ -285,6 +285,12 @@ func withTo(b *builders.RuleBuilder, hosts []string, rule gatewayv2alpha1.Rule)
 }
 
 func withFrom(b *builders.RuleBuilder, rule gatewayv2alpha1.Rule, oryPassthrough bool) *builders.RuleBuilder {
+	// only viable when migration step is happening. Do not add ingressgateway source during migration
+	if rule.Jwt != nil && oryPassthrough {
+		return b.WithFrom(builders.NewFromBuilder().
+			WithForcedJWTAuthorizationV2alpha1(rule.Jwt.Authentications).
+			Get())
+	}
```
barchw, line 293: Why is this change needed, as L307-311 exists?
https://github.com/kyma-project/api-gateway/blob/2158cff5a67f56f7881904e6cc0b2e1709eb824d/internal/processing/processors/v2alpha1/authorizationpolicy/creator.go#L307-L311

barchw (changes requested):

internal/processing/processors/v2alpha1/authorizationpolicy/creator.go
barchw: Ah, I understand know, for JWT this was exiting before adding the from oathkeeper principal.
I guess better solution would be to have that piece of code before returning.

barchw: After consideration, we will keep this in the way it is, as there might be side effects of forcing oathkeeper as the from.

barchw (approved):

PR #1922, opened 2025-05-13
Adjust "needs" statement

kasiakepka:
**Description**

- Adjust "needs" statement

### DIFF
--- a/.github/workflows/release-api-gateway-2.yaml
+++ b/.github/workflows/release-api-gateway-2.yaml
@@ -62,7 +62,7 @@ jobs:
     uses: ./.github/workflows/call-unit-lint.yaml
 
   integration-tests:
-    needs: build-image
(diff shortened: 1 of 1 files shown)


.github/workflows/release-api-gateway-2.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
     uses: ./.github/workflows/call-unit-lint.yaml
 
   integration-tests:
-    needs: build-image
+    needs: [build-image, check-prerequisites]
```
barchw, line 65: As build-image already requires check-prerequisites we don't need to add it in here

barchw: OK, It's needed for the needs below 🆗

barchw (approved):

Issue #1988, opened 2025-06-04
Update tests to use APIRule v2


strekm:
**Description**

Adapt all tests testing in any way APIRule v1beta1 to use APIRule v2 in original version v1beta1. Evaluate updating tests directly to version v2 if it makes sense. Tests that need to stay in v1 are ones focusing on migration and on testing reconciliation of v1. All other tests can be updated to use v2. 

**ToDos [Developer]**

- [x] make sure original version is applied accordingly when use v2 with v1beta1-spec annotation
- [x] switch custom domain tests to v2
- [x]  adapt v1 tests that use not supported handlers to use v2 with v1beta1-spec annotation (and original version v1beta1)
- [x]  change all v2alpha1 usages to v2
- [x] ory tests use v2 with v1beta1-spec annotation (and original version v1beta1)

**PRs**

- [ ] [...]()

**ACs [PO]**

- [x] migration from v1 to v2 is tested
- [x] v1 reconciliation is tested
- [x] all other tests updated to v2

PR #2010, opened 2025-06-16
bug: fix assignment of LastProcessedTime field in apirule_controller.go

Ressetkk:
* fix assignment of LastProcessedTime field in apirule_controller.go
* pin curl to specific version - 8.14.1 has small regression https://github.com/curl/curl/issues/17554
* adjust upgrade scenario to only use v2(alpha1)

#1988

### DIFF
--- a/controllers/gateway/apirule_controller.go
+++ b/controllers/gateway/apirule_controller.go
@@ -113,16 +113,16 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		l.Error(err, "Error while getting APIRule v2alpha1")
 		return doneReconcileErrorRequeue(err, errorReconciliationPeriod)
 	}
+
+	// assign LastProcessedTime early to indicate that resource got reconciled
+	apiRuleV2alpha1.Status.LastProcessedTime = metav1.Now()
+
 	apiRule := gatewayv1beta1.APIRule{}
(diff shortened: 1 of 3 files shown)


controllers/gateway/apirule_controller.go
```diff
@@ -113,16 +113,17 @@ func (r *APIRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		l.Error(err, "Error while getting APIRule v2alpha1")
 		return doneReconcileErrorRequeue(err, errorReconciliationPeriod)
 	}
+
+	// assign LastProcessedTime and ObservedGeneration early to indicate that
+	// resource got reconciled
```
kubicar, line 118:
```suggestion
	// assign LastProcessedTime early to indicate that resource got reconciled
```

config/crd/bases/gateway.kyma-project.io_apirules.yaml
```diff
@@ -232,10 +232,6 @@ spec:
                 maximum: 3900
                 minimum: 1
                 type: integer
-            required:
```
kubicar, line 235: i am not sure that this is good move, I know that changes itself was introduced previous but we need to consider this validation. that will create a possibility to apply now valid ApiRule

kubicar (approved):

PR #2064, opened 2025-07-14
fix: local rate limiting targeting Istio ingress gateway

kubicar:
**Description**

- ...

**Pre-Merge Checklist**

**Related issues**


tests/integration/pkg/hooks/api-gateway.go
```diff
@@ -345,6 +347,22 @@ func deleteBlockingResources(ctx context.Context) error {
 		}
 	}
 
+	rateLimitList := ratelimit.RateLimitList{}
+	err = k8sClient.List(ctx, &rateLimitList)
+	if err == nil {
```
kolodziejczak, line 352: I'd say it's a nit in the context of post hook in the e2e tests, but we don't handle error here. It won't have any bad effect on us here, but maybe it would be worth to log it out at least if the error occurs, so we can catch any hidden issues with resource pruning?

kolodziejczak (changes requested):

tests/integration/pkg/hooks/api-gateway.go
kubicar: It is just pruning, error which we can spot in that case is only broken communication with server. IMHO It can be solved in new e2e tests. But I will fix it, it's a total basic to give an information when something was failed.

kolodziejczak (approved):

docs/user/custom-resources/ratelimit/04-00-ratelimit.md
```diff
@@ -13,21 +13,23 @@ kubectl get crd ratelimits.gateway.kyma-project.io -o yaml
...
 | **enableResponseHeaders**             | **NO**   | Enables **x-rate-limit** response headers. The default value is `false`.                                                                                                                                                                                                                                                                                                                           |
 | **enforce**                           | **NO**   | Specifies whether the rate limit should be enforced. The default value is `true`.                                                                                                                                                                                                                                                                                                                  |
 
+## RateLimit for Istio Ingress Gateway
+To rate limit requests to the Istio Ingress Gateway, you must create a RateLimit Custom Resource in the `istio-system` namespace and set `selectorLabels` field to point Istio Ingress Gateway `app: istio-ingressgateway`.
```
nataliasitko, line 32:
```suggestion
To rate limit requests to the Istio Ingress Gateway, you must create a RateLimit custom resource in the `istio-system` namespace and set the **selectorLabels** field to point to the Istio Ingress Gateway by including the label `app: istio-ingressgateway`.
```

kolodziejczak (approved):

nataliasitko (approved):

PR #2082, opened 2025-07-21
JWT test using the new approach

barchw:
**Description**

- JWT test using the new approach

**Pre-Merge Checklist**

**Related issues**


tests/e2e/pkg/helpers/httpbin/httpbin.go
```diff
@@ -0,0 +1,87 @@
...
+		opt(opts)
+	}
+
+	r, err := infrahelpers.ResourcesClient(t)
+	if err != nil {
+		t.Logf("Failed to get resources client: %v", err)
+		return "", 0, fmt.Errorf("failed to get resources client: %w", err)
+	}
+
+	err = infrahelpers.CreateNamespace(t, opts.Namespace, infrahelpers.IgnoreAlreadyExists(), infrahelpers.WithSidecarInjectionEnabled())
```
mluk-sap, line 46: This is little bit unexpected. It should rather deploy it in the existing namespace...

tests/e2e/pkg/setup/cleanup.go
```diff
@@ -0,0 +1,35 @@
+package setup
+
+import (
+	"context"
+	"os"
+	"testing"
+)
+
+var forceSkipCleanup = os.Getenv("FORCE_SKIP_CLEANUP") == "true"
```
mluk-sap, line 9: This is not intuitive...
Furthermore, I'm not sure if it makes sense.

Cleanup is rather required to make multiple tests working correctly in order to not interfere and don't use too many resources. In other words - disabling cleanup would leave hundreds of apirules and dozens of namespaces with dozens of httpbin resources.

The intention behind the 'Skip cleanup' allows to leave everything as is in case of failure, so the resources are not mutated or deleted, which allows manual exploration to see what happened. But in this case we would see only resources used by this single failed test. Resources created by previous tests should have been cleaned before.

tests/e2e/pkg/helpers/testsetup/testsetup.go
```diff
@@ -0,0 +1,79 @@
...
+	Namespace         string
+	TargetServiceName string
+	TargetServicePort int
+	Mock              *oauth2mock.Mock
+}
+
+// SetupRandomNamespaceWithOauth2MockAndHttpbin creates a namespace with a random ID,
+// deploys the oauth2mock and httpbin services, and returns a TestBackground struct.
+// It also sets up the namespace with sidecar injection enabled.
+func SetupRandomNamespaceWithOauth2MockAndHttpbin(t *testing.T, options ...Option) (bckg TestBackground, err error) {
```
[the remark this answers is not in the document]
barchw: I will hide that behind an interface

kubicar (approved):

nataliasitko (approved):

PR #2099, opened 2025-07-28
Combine OS release and module-manifest release

kasiakepka:
**Description**

- Add module-manifests release to the OS release process

### DIFF
--- a/.github/workflows/release-api-gateway-2.yaml
+++ b/.github/workflows/release-api-gateway-2.yaml
@@ -249,7 +249,7 @@ jobs:
       - name: Setup git and clone repo
         shell: bash
         run: |
-          git config --global http.{{ secrets.INTERNAL_GITHUB_URL }}.version "HTTP/1.1"
+          git config --global http.${{ secrets.INTERNAL_GITHUB_URL }}.version "HTTP/1.1"
           git config --global user.email ${{ vars.GIT_USER_EMAIL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-api-gateway-2.yaml
```diff
@@ -198,3 +198,123 @@ jobs:
...
+              exit 1  
+          fi
+      - name: Set branch name
+        shell: bash
+        run: echo "BRANCH_NAME=api-gateway-${TAG}" >> $GITHUB_ENV
+
+      - name: Setup git and clone repo
+        shell: bash
+        run: |
+          git config --global http.https://github.tools.sap.version "HTTP/1.1"
```
kolodziejczak, line 252: We can't have this url publicly visible

kolodziejczak (changes requested):
I don't like the idea of pushing anything from public repos to private ones. I think we should definitely drop this direction of pushes. We should have an internal workflow that reacts on public releases, and pulls it into the internal instead of public workflow pushing it there.

kasiakepka:
> I don't like the idea of pushing anything from public repos to private ones. I think we should definitely drop this direction of pushes. We should have an internal workflow that reacts on public releases, and pulls it into the internal instead of public workflow pushing it there.

Fully agree, I think we already have a task to try out this approach. What I am doing here is just adjustment of already existing process., migrating to the internal repo sounds like a whole new topic.

PR #2152, opened 2025-08-21
Admission webhook preventing CREATE and UPDATE of APIRule in version …

barchw:
**Description**

- ...

**Pre-Merge Checklist**

**Related issues**


apis/gateway/v1beta1/apirule_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+
+// +kubebuilder:webhook:path=/validate-gateway-kyma-project-io-v1beta1-apirule,mutating=false,failurePolicy=fail,sideEffects=None,groups=gateway.kyma-project.io,resources=apirules,verbs=create;update;delete,versions=v1beta1,name=v1beta1-admission.apirule.gateway.kyma-project.io,admissionReviewVersions=v1,servicePort=9443,serviceName=api-gateway-webhook-service,serviceNamespace=kyma-system,matchPolicy=Exact
+// +kubebuilder:object:generate=false
+type ValidatingWebhook struct {
+	Client client.Client
+}
+
+func (w *ValidatingWebhook) shouldBlockAPIRule() bool {
+	configMap := &corev1.ConfigMap{}
+	err := w.Client.Get(context.Background(), types.NamespacedName{
```
kubicar, line 80: Optimization concern: potential latency (don't remember if this resource is cached in client, I will check it). We want to make a call each time someone will use webhook? IMHO i can be set during setup, in this way it will not make additional latency on webhook execution, but removal or create of CM will require restart of operator (or aditional logic to observe CM)

kubicar: It is only on modyfication actions, it will not harm performance.

kubicar (approved):

PR #1987, opened 2025-06-04
chore: cleanup tests

Ressetkk:
With this PR there no v1beta1 APIRule manifests in `./tests/integration/` directory.

- istio-jwt: convert v1beta1 APIRule manifests to v2
- istio-jwt: upgrade v2alpha1 scenario to v2
- istio-jwt: de-template some unnecessary manifests
- istio-jwt: clean up empty steps
- istio-jwt: move some functions to appropriate locations
- ory: convert v1beta1 APIRule manifests to v2
- ory: clean up empty steps
- ory: remove v2alpha1 functions in favour of generic approach
- gateway: convert v1beta1 APIRule manifests to v2
- api_rule helpers: remove duplicate code and refactor to use APIRuleStatus interface in tests
- custom-domain: upgrade manifests from v1beta1 to v2
- custom-domain: re-implement oauth2 introspection scenario to use v2-compatible JWT
- apirule: reconcile: refactor `originalversionchanged` predicate to re-use and include changes in `v1beta1-spec` annotation
- apirule: conversion: do not replace `original-version` annotation if the spec is empty (incompatible apirule case)

#1988

PR #2157, opened 2025-08-22
Remove required fields in test code

barchw:
**Description**

- Remove git patch for allowing v2 without particular fields in tests and do it via Kube API

**Pre-Merge Checklist**

**Related issues**


tests/integration/testsuites/patch/patch.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Apply the patch to the APIRule CRD
+	return k8sClient.Patch(context.Background(), &apiextensionsv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "apirules.gateway.kyma-project.io",
+		},
+	}, client.RawPatch(types.JSONPatchType, patchData))
+}
+
+func Removev2alpha1VersionRequiredFields(k8sClient client.Client) error {
```
kubicar, line 43: Are you sure that take of restriction from fields is best solution. What with idea that we just serve APIRule v1beta1 -> apply resource -> userve it in e2e test as steps in feature files?

barchw: I'm pretty sure doing the above would invalidate the already done switch from using v1beta1 API to v2
This patch is required for the changes that were done in https://github.com/kyma-project/api-gateway/pull/1987
To add to that, I think it is better that we require the usage of v2

kubicar (approved):

PR #2225, opened 2025-09-10
tests: e2e rewrite ext-auth scenario to new approach

kubicar:
**Description**

- New implementation of e2e tests for extauth

**Pre-Merge Checklist**

- [ ] ~The code coverage is acceptable.~
- [ ] ~Release notes for the introduced changes are created.~
- [ ] ~If Kubebuilder changes were made, you ran `make generate-manifests` and committed the changes before the merge.~
- [ ] ~Pre-existing managed resources are correctly handled.~
- [ ] ~The change works on all hyperscalers supported by SAP BTP, Kyma runtime.~
- [ ] ~There is no upgrade downtime.~
- [ ] ~For infrastructure changes, you checked if the changes affect the hyperscaler's costs.~
- [ ] ~RBAC settings are as restrictive as possible.~
- [ ] ~If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.~
- [ ] ~You checked if this change should be cherry-picked to active release branches.~
- [ ] ~The configuration does not introduce any additional latency.~

**Related issues**

### DIFF
--- a/.github/actions/e2e-k3d/action.yaml
+++ b/.github/actions/e2e-k3d/action.yaml
@@ -22,16 +22,10 @@ runs:
     - uses: actions/setup-go@v5
       with:
         go-version-file: "go.mod"
-    - name: Checkout to PR branch # to remove after getting rid of pull_request_target
-      shell: bash
-      if: github.event_name == 'pull_request' || github.event_name == 'pull_request_target'
-      run: |
-        git fetch origin pull/${{ github.event.number }}/head:PR-${{ github.event.number }}
-        git checkout PR-${{ github.event.number }}
     - name: Create Cluster
       uses: ./.github/actions/provision-k3d-cluster
       with:
-          version: "1.32.6"
+          version: "1.33.0"
     - name: Run integration tests
       shell: bash
       env:
--- a/.github/actions/e2e-k3d/action.yaml
+++ b/.github/actions/e2e-k3d/action.yaml
@@ -1,0 +1,85 @@
+name: 'E2E (k3d)'
+description: 'Runs integration tests on k3d'
+inputs:
+  manager_image:
+    description: 'Manager image used for test'
+    required: true
+  test_make_target:
+    description: 'Make target for integration tests to run'
+    required: true
+  test_client_id:
+    description: "Client id for oidc"
+    required: false
+  test_client_secret:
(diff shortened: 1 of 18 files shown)


tests/e2e/pkg/helpers/modules/istio.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		t.Logf("Failed to create Istio custom resource: %v", err)
+		return err
+	}
+
+	setup.DeclareCleanup(t, func() {
+		t.Log("Cleaning up Istio after the tests")
+		err := TeardownIstioCR(t, options...)
+		if err != nil {
+			t.Logf("Failed to clean up APIGateway custom resource: %v", err)
```
kolodziejczak, line 74: I think the log should contain `Istio custom resource`

tests/e2e/pkg/helpers/modules/istio.go
```diff
@@ -0,0 +1,137 @@
...
+		} else {
+			t.Log("APIGateway custom resource cleaned up successfully")
```
kolodziejczak, line 76: I think the log should contain `Istio custom resource`

tests/e2e/tests/extauth/ext_auth_apirule.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: gateway.kyma-project.io/v2
+kind: APIRule
+metadata:
+  name: "{{ .Name }}"
+spec:
+
```
kolodziejczak, line 6: just a nit, but let's remove this empty line

.github/actions/e2e-k3d/action.yaml
kolodziejczak, line 1: As I understand, we are introducing new action because of the hardcoded way of deploying Istio module. Also all of the jobs eventually will require this new way of deploying Istio module, as the IstioCR in the new tests is created from the code. What do you think, maybe it would be worth to create some comment or follow-up to not to forget that after rewriting all tests, we can remove the old action?

kolodziejczak (changes requested):
Do we also need to change other jobs on different targets? like post merge or release?

kubicar:
for now we are not adding other target workflows

PR #2266, opened 2025-09-30
Dump logs and resources

barchw:
**Description**

- Dump pod logs and resources 

**Pre-Merge Checklist**

- [ ] The code coverage is acceptable.
- [ ] Release notes for the introduced changes are created.
- [ ] If Kubebuilder changes were made, you ran `make generate-manifests` and committed the changes before the merge.
- [ ] Pre-existing managed resources are correctly handled.
- [ ] The change works on all hyperscalers supported by SAP BTP, Kyma runtime.
- [ ] There is no upgrade downtime.
- [ ] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [ ] RBAC settings are as restrictive as possible.
- [ ] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [ ] You checked if this change should be cherry-picked to active release branches.
- [ ] The configuration does not introduce any additional latency.

**Related issues**


tests/e2e/pkg/helpers/oauth2/mock/manifest.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
         - name: mock-oauth2-server
-          image: europe-central2-docker.pkg.dev/sap-se-cx-kyma-goat/networking-dev-tools/oauth2-mock:latest
+          image: i566483/mock-oauth2:1
```
mluk-sap, line 18: Is this intended? :)

tests/e2e/pkg/setup/dump.go
```diff
@@ -0,0 +1,177 @@
...
+
+			err = fileHandle.Close()
+			if err != nil {
+				t.Logf("Could not close log file: err=%s", err)
+			}
+		}
+	}
+}
+
+func getLogsFromAllPods(t *testing.T) {
```
mluk-sap, line 101: The function shouldn't be called 'get' as it doesn't get anything, we may call it 'storeLogsFromAllPods' instead

tests/e2e/pkg/setup/dump.go
```diff
@@ -0,0 +1,177 @@
...
+	"k8s.io/client-go/discovery"
+	"log"
+	"os"
+	"path"
+	"sigs.k8s.io/yaml"
+	"testing"
+)
+
+const (
+	podLogsDir  = "/tmp/logs/%s/pods"
```
mluk-sap, line 20: Storing logs in /tmp dir is a bad practice:
- it doesn't honor the [TMPDIR](https://en.wikipedia.org/wiki/TMPDIR) env variable, so it won't work if /tmp doesn't exist
- this would not work in non-Unix operating systems
- the subpath (/logs) is also hardcoded, so we'll mix logs from multiple executions

To avoid above problems the standard libraries in programing languages offer functionalities like: https://pkg.go.dev/os#CreateTemp (which uses https://pkg.go.dev/os#TempDir)

Furthermore the 'tmp' directory is intended to store data only for temporary usage, so files should be deleted immediately after they are not needed anymore (they've been already 'processed').

Here the use case if slightly different - we want to store them for a longer time, which is beyond the 'go test' runtime:
- so the pipeline can take it and archive it
- the end-developer may look at them

Thus, I would rather recommend to store it either somewhere inside the 'current dir' or inside the api-gateway root dir.

tests/e2e/pkg/setup/dump.go
```diff
@@ -0,0 +1,177 @@
...
+	podLogsPath = "/tmp/logs/%s/pods/%s.%s.log"
+	logDir      = "/tmp/logs/%s"
+)
+
+func DumpClusterResources(t *testing.T) {
+	t.Helper()
+	dumpPath := path.Join(fmt.Sprintf(logDir, t.Name()), "resources")
+	_, err := os.Stat(dumpPath)
+	if !os.IsNotExist(err) {
```
mluk-sap, line 29: I'm thinking whether this is good approach. It ensures that the directory is created once, which is okay, but if for whatever reason (pipeline issue, developer doing mistake) the test is executed twice then we would have logs only from the first run, which may be very misleading.

Maybe it would be better to ensure that DumpClusterResources is called once inside the DeclareCleanup implementation and then consequently fail-fast if the directory already pre-exists.

Another idea is to create separate directories for each 'go test' run. The directory name can be generated upon startup (based on timestamp) and stored in some package-scoped variable.

mluk-sap: Okay, we agreed to include timestamp in the directory name

mluk-sap (approved):

PR #2623, opened 2026-04-24
bug: apirule: filter events by event type

Ressetkk:
introduce helper function to filter events by event type

controller-runtime during cold-boots initializes waits for the cache to be done. With many-to-many constraints, where we have to list all apirules and match the service to it, querying APIServer takes a lot of time on large clusters.

As a workaround, we filter out CREATE events, so we only care about the Service update, not creation.

This is needed, because on cold-startup controller waits for all events to be preprocessed before readiness. Since we make API call on every event, with 20k+ resources, we need around 30 minutes to initialize controller. With overloaded apiserver that adds 70-200ms of delay between events. When we ignore CREATE events, we don't query problematic List() calls on startup.

This issue does not affect the same Service watcher in apigateway controller, because there we have a specific breakpoints based on service name and namespace, and API call is made only *once*.

### DIFF
--- a/docs/release-notes/3.7.0.md
+++ b/docs/release-notes/3.7.0.md
@@ -1,2 +1,5 @@
 ## New Features
 - DNSEntry custom resources now support the DualStack IP stack. The operator detects the IP stack of the `istio-ingressgateway` Service by inspecting the **spec.ipFamilies** field and sets the appropriate annotation on the DNSEntry resource. See [issue #2587](https://github.com/kyma-project/api-gateway/issues/2587).
+
+## Bug Fixes
+- Fixed slow controller startup by filtering out Service CREATE events in the Service watcher. Previously, the controller queried the API server for every event during cache initialization, causing prolonged startup times. See [PR #2623](https://github.com/kyma-project/api-gateway/pull/2623).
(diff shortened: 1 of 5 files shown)


mluk-sap (dismissed):

docs/release-notes/3.7.0.md
```diff
@@ -1,2 +1,5 @@
 ## New Features
 - DNSEntry custom resources now support the DualStack IP stack. The operator detects the IP stack of the `istio-ingressgateway` Service by inspecting the **spec.ipFamilies** field and sets the appropriate annotation on the DNSEntry resource. See [issue #2587](https://github.com/kyma-project/api-gateway/issues/2587).
+
+## Bug Fixes
+- Fixed slow controller startup by filtering out Service CREATE events in the Service watcher. Previously, the controller queried the API server for every event during cache initialization, causing prolonged startup times. See [PR #2623](https://github.com/kyma-project/api-gateway/pull/2623).
```
nataliasitko:
```suggestion
- The controller now filters out Service CREATE events in the Service watcher instead of querying the API server for every event during cache initialization. The new behavior reduces startup time. See [PR #2623](https://github.com/kyma-project/api-gateway/pull/2623).
```

joanbod (approved):

nataliasitko (approved):

PR #2769, opened 2026-07-28
feat(test): added dualstack tests + possibility to run tests on gardener

kubicar:
**Description**

- ...

**Pre-Merge Checklist**

- [x] The code coverage is acceptable.
- [x] Release notes for the introduced changes are created.
- [x] If Kubebuilder changes were made, you ran `make manifests` and committed the changes before the merge.
- [x] Pre-existing managed resources are correctly handled.
- [x] The change works on all hyperscalers supported by SAP BTP, Kyma runtime.
- [x] There is no upgrade downtime.
- [x] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [x] RBAC settings are as restrictive as possible.
- [x] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [x] You checked if this change should be cherry-picked to active release branches.
- [x] The configuration does not introduce any additional latency.
- [x] You checked if Busola updates are needed.

**Related issues**


kubicar:
https://github.com/kyma-project/istio/issues/2139

hack/ci/install-dualstack-prereqs.sh
```diff
@@ -0,0 +1,46 @@
...
+if [[ "$ISTIO_MANAGER_VERSION" == "latest" ]]; then
+  ISTIO_MANAGER_URL="https://github.com/kyma-project/istio/releases/latest/download/istio-manager-experimental.yaml"
+else
+  ISTIO_MANAGER_URL="https://github.com/kyma-project/istio/releases/download/${ISTIO_MANAGER_VERSION}/istio-manager-experimental.yaml"
+fi
+
+kubectl create namespace kyma-system --dry-run=client -o yaml | kubectl apply -f -
+kubectl label namespace kyma-system istio-injection=enabled --overwrite
+
+printf 'networkDetails:\n  dualStackIPEnabled: true\n' \
```
mluk-sap, line 36: This would be more readable:
```
kubectl create configmap -n kyma-system kyma-provisioning-info --from-file=details=/dev/stdin --dry-run=client -o yaml << 'EOF' | kubectl apply -f -
networkDetails:
  dualStackIPEnabled: true
EOF
```

hack/ci/install-dualstack-prereqs.sh
```diff
@@ -0,0 +1,46 @@
...
+  | kubectl create configmap -n kyma-system kyma-provisioning-info \
+      --from-file=details=/dev/stdin \
+      --dry-run=client -o yaml \
+  | kubectl apply -f -
+
+tmp="$(mktemp -t istio-manager-experimental.XXXXXX.yaml)"
```
mluk-sap, line 42: We usually do it via kubectl apply -f:
```
kubectl apply -f https://github.com/kyma-project/istio/releases/latest/download/istio-manager-experimental.yaml
```

hack/ci/install-dualstack-prereqs.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Idempotent. Safe to re-run.
+#
+# Uses whatever KUBECONFIG the caller has set.
+
+set -euo pipefail
+
+ISTIO_MANAGER_VERSION="${ISTIO_MANAGER_VERSION:-latest}"
+if [[ "$ISTIO_MANAGER_VERSION" == "latest" ]]; then
+  ISTIO_MANAGER_URL="https://github.com/kyma-project/istio/releases/latest/download/istio-manager-experimental.yaml"
```
mluk-sap, line 28: I think that we have an asymmetry here:
- in normal (single-stack) tests the module (Istio in this case) is installed via make deploy command inside hack/ci/integration-test-gardener.sh; moreover, it uses kustomize, so the deployment manifest is generated
- here (in dual stack case), we have an additional script that does it differently (installs file from the repository)
I think that we should unify this.
In particular - kyma-provisioning-info configmap may be always installed and the value of dualStackIPEnabled may be taken from the TEST_IP_FAMILY

hack/ci/set-aws-dualstack-gardener.sh
```diff
@@ -0,0 +1,14 @@
...
+# Auto-loaded by provision-gardener.sh / deprovision-gardener.sh when
+# GARDENER_PROVIDER=aws-dualstack. Shoot template lives in
+# shoot_aws-dualstack.yaml and enables VPC dual-stack + AWS LBC.
+
+MACHINE_TYPE="m5.xlarge"
+DISK_SIZE=50
+DISK_TYPE="gp2"
+SCALER_MAX=3
+SCALER_MIN=3 # 3 machines to spread across 3 zones with HA control plane
+GARDENER_PROVIDER="aws-dualstack"
```
mluk-sap, line 11: We are mixing two things here:
- the provider (aws/gcp/azure, etc.)
- cluster configuration/flavour (dual stack)
I think we should:
- either split it into multiple variables (GARDENER_PROVIDER / GARDENER_FLAVOUR)
- or rename it, so it is clear that it represents the configuration (GARDENER_CONFIGURATION)

Makefile
```diff
@@ -214,6 +214,32 @@ e2e-test-part1:
...
+dualstack-e2e-test-part1: ## Dualstack subset, part 1: LB-dialling suites only (asterisk, cors, expose_methods_on_paths, extgateway). Non-network suites (request, service, validation) are excluded — the family axis is inert for them.
+	make -C tests/e2e/tests dualstack-e2e-test-part1
+
+.PHONY: dualstack-e2e-test-part2
+dualstack-e2e-test-part2: ## Dualstack subset, part 2: auth + in-cluster HTTP suites (extauth, jwt, no_auth, short_host).
+	make -C tests/e2e/tests dualstack-e2e-test-part2
+
+.PHONY: install-dualstack-prereqs
+install-dualstack-prereqs: ## Install the prerequisites the tests/e2e/tests suites need to exercise dualstack on a Gardener AWS shoot (labels kyma-system for istio injection, applies the kyma-provisioning-info CM that gates dualstack in istio-manager, installs the experimental istio-manager which is the only build that honours dualStackIPEnabled). See hack/ci/install-dualstack-prereqs.sh for the rationale.
+	./hack/ci/install-dualstack-prereqs.sh
```
mluk-sap, line 227: With that we introduce mutual coupling between Makefile and hack/ci scripts. Usually hack/ci scripts call Makefile targets (like make deploy, make install-istio, etc.), while in the Makefile we don't call scripts, especially from hack/ci directory.

So in other words, we have these layers:
- golang code
- Makefile - which contains simple shell commands, like go, kustomize, controller-gen, mainly to provide parameters
- shell scripts in hack/ci - these orchestrate bigger processes, like cluster provisioning, preparation for tests (module installation, etc.)

Moreover, the kyma-provisioning-info configmap installation belongs to the provisioning process. It should be executed only once when the cluster is installed.

So the ideal place for it is here: https://github.com/kyma-project/api-gateway/blob/main/hack/ci/provision-gardener.sh#L119

mluk-sap (changes requested):
Okay, we discussed it with @kubicar.
The idea is to remove install-prereqs script and install the kyma-provisioning-info configmap in the script that provisions the Gardener cluster (as it belongs to provisioning).
With that change we may simplify the Makefile (remove newly added targets) and we may simplify the CI setup, because we may reuse existing github action that runs e2e tests and add ip_families as a parameter.
Additionally, we'd like to introduce a GARDENER_CONFIGURATION env var, which will be responsible for choosing proper directory with gardener configuration (vars.sh, shoot.yaml). The GARDENER_PROVIDER and DUAL_STACK/IP_FAMILIES would then be set by the vars.sh script (renamed existing set-*-gardener.sh script).

.github/actions/e2e-test-gardener-dualstack/action.yaml
mluk-sap, line 1: We may remove this file.
The existing action e2e-test-gardener may handle it. We only need to add test_ip_family parameter to it.
The dualstack-e2e-setup make target is also not required, because:
- kyma-provisioning-info configmap should be created in the script that performs cluster provisioning: https://github.com/kyma-project/api-gateway/blob/9712b7cfe6a1b96c9f9b8f87493246a253ca9469/hack/ci/provision-gardener.sh#L119
- Istio experimental should be deployed in the script that prepares cluster for tests: https://github.com/kyma-project/api-gateway/blob/9712b7cfe6a1b96c9f9b8f87493246a253ca9469/hack/ci/integration-test-gardener.sh#L63
Then everything should work as before, so the e2e-test-gardener Github Action should be able to run these tests via existing Makefile targets

hack/ci/install-dualstack-prereqs.sh
mluk-sap, line 1: We may remove this file as explained before:
- kyma-provisioning-info configmap should be created in the script that performs cluster provisioning: https://github.com/kyma-project/api-gateway/blob/9712b7cfe6a1b96c9f9b8f87493246a253ca9469/hack/ci/provision-gardener.sh#L119
- Istio experimental should be deployed in the script that prepares cluster for tests: https://github.com/kyma-project/api-gateway/blob/9712b7cfe6a1b96c9f9b8f87493246a253ca9469/hack/ci/integration-test-gardener.sh#L63

hack/ci/install-dualstack-prereqs.sh
mluk-sap: We discussed that, it would be moved to the script that prepares cluster for tests:
https://github.com/kyma-project/api-gateway/blob/9712b7cfe6a1b96c9f9b8f87493246a253ca9469/hack/ci/integration-test-gardener.sh#L63

hack/ci/set-aws-dualstack-gardener.sh
mluk-sap: We discussed that, we'd like to:
- introduce GARDENER_CONFIGURATION variable that would be responsible for choosing the directory with the given Gardener configuration
- every such directory would contain vars.sh (renamed set-*-gardener.sh) and shoot.yaml
- vars.sh would contain:
  - GARDENER_PROVIDER (aws/gcp/azure/etc.)
  - IP_FAMILY (or equivalent)

With this approach it is more generic, the job that orchestrates the test uses only one variable to define the particular configuration. With this approach we may introduce even more factors in future, as the job would be only aware of the 'name', so it is not limited to the provider and dual stack only.

Makefile
mluk-sap: We discussed it and we'd like to remove this and move:
- kyma-provisioning-info configmap creation to the script that provisions the cluster
- Istio experimental installation to the script that prepares cluster for tests (it already calls make install-istio)

Makefile
```diff
@@ -214,6 +214,32 @@ e2e-test-part1:
 e2e-test-part2:
 	make -C tests/e2e/tests e2e-test-part2
 
+.PHONY: dualstack-e2e-test-part1
```
mluk-sap, line 217: We should do it in a generic way using [Pattern rule](https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html#Pattern-Rules) to avoid adding the same PHONY targets to both places.

See how we did it in the Istio: https://github.com/kyma-project/istio/blob/4fc775de1544bdb7dff8e43c3f04e35d28b5a4e9/Makefile#L231-L236

hack/ci/gardener/provision.sh
```diff
@@ -125,5 +124,18 @@ kubectl create  --kubeconfig "${GARDENER_KUBECONFIG}" \
     jq -r ".status.kubeconfig" | \
     base64 -d > "${CLUSTER_KUBECONFIG}"
 
+# The kyma-provisioning-info ConfigMap gates dualstack in istio-manager. On real
+# BTP clusters infrastructure-manager writes it; Gardener-native CI shoots must
+# fill it in themselves. It belongs to provisioning, so it is created once here,
+# and dualStackIPEnabled follows the shoot's IP stack (dualstack shoots enable it).
+echo "Creating kyma-provisioning-info ConfigMap"
+[ "${GARDENER_IP_STACK}" = "dualstack" ] && dual_stack_enabled=true || dual_stack_enabled=false
+kubectl --kubeconfig "${CLUSTER_KUBECONFIG}" create namespace kyma-system \
```
mluk-sap: Let's align this with the Istio repo and move it to the Makefile

Makefile
mluk-sap: Ok, we decided to not do it this way. However, in the Istio repository we use w different approach and we don't have separate Makefile in the tests/e2e/tests subdirectory. We'll align it in the future.

Makefile
```diff
@@ -219,6 +235,18 @@ e2e-test-custom-domain:
 e2e-test-gateway-gardener:
 	make -C tests/e2e/tests e2e-test-gateway-gardener
 
+.PHONY: dualstack-e2e-test-part1
+dualstack-e2e-test-part1: ## Dualstack subset, part 1: LB-dialling suites only (asterisk, cors, expose_methods_on_paths, extgateway). Non-network suites (request, service, validation) are excluded — the family axis is inert for them.
+	make -C tests/e2e/tests dualstack-e2e-test-part1
```
mluk-sap: Doesn't work:
```
make dualstack-e2e-test-part1
make -C tests/e2e/tests dualstack-e2e-test-part1
make[1]: *** No rule to make target `dualstack-e2e-test-part1'.  Stop.
make: *** [dualstack-e2e-test-part1] Error 2
```

mluk-sap (approved):

PR #2800, opened 2026-08-17
Adapt status according to ADR

TomaszSkrzyp:
**Description**

- **feature: added conditional "Processing" status setting. The conditional status setting based partly on CR "generation" value**
- **We set the status either when the APIGateway CR is being installed or when the APIGateway CR spec changed, detected based on generation change**
- **fix: added tests and improved generation bump logic**

**Pre-Merge Checklist**

- [ ] The code coverage is acceptable.
- [ ] Release notes for the introduced changes are created.
- [ ] If Kubebuilder changes were made, you ran `make manifests` and committed the changes before the merge.
- [ ] Pre-existing managed resources are correctly handled.
- [ ] The change works on all hyperscalers supported by SAP BTP, Kyma runtime.
- [ ] There is no upgrade downtime.
- [ ] For infrastructure changes, you checked if the changes affect the hyperscaler's costs.
- [ ] RBAC settings are as restrictive as possible.
- [ ] If any new libraries are added, you verified license compliance and maintainability, and made a comment in the PR with details. We only allow stable releases to be included in the project.
- [ ] You checked if this change should be cherry-picked to active release branches.
- [ ] The configuration does not introduce any additional latency.
- [ ] You checked if Busola updates are needed.

**Related issues**


Ressetkk:
The PR implements incorrect design. The one you should pay attention is https://github.com/kyma-project/api-gateway/blob/main/docs/contributor/adr/0011-module-status.md. We should avoid overusing `Processing` state, and ideally set it *only* when the status of the module was not known (e.g. empty Status field).

docs/release-notes/3.11.0.md
```diff
@@ -0,0 +1,5 @@
+## Processing Status Limited to Spec Changes
+
+Previously, the `Processing` status was set on every reconciliation of the APIGateway CR, including periodic re-syncs with no user-initiated changes. To reduce unnecessary status transitions, the `Processing` status is now only set when the CR generation advances — that is, when the spec actually changes — or on initial installation.
+
```
Ressetkk: I think you can remove the blank line between these sentences. They don't need to be separated.

internal/controller/operator/processing_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/go-logr/logr"
+	operatorv1alpha1 "github.com/kyma-project/api-gateway/apis/operator/v1alpha1"
+	"github.com/kyma-project/api-gateway/internal/conditions"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var _ = Describe("shouldSetProcessing", func() {
```
Ressetkk: Testing private functions is pointless. If you can, you should test this code from the controller code to make sure the Reconcile() call reaches this code as expected.

internal/controller/status.go
```diff
@@ -200,6 +201,23 @@ func UpdateApiGatewayStatus(ctx context.Context, k8sClient client.Client, apiGat
...
+				prevByType[c.Type] = c
+			}
+			for i := range newStatus.Conditions {
+				//while Processing, do not advance generation, preserve previous value if present.
+				if prev, ok := prevByType[newStatus.Conditions[i].Type]; ok {
+					newStatus.Conditions[i].ObservedGeneration = prev.ObservedGeneration
+				}
+			}
+		}
+
```
Ressetkk: You should avoid using `else` statements and return early to reduce code coupling.

internal/controller/status.go
```diff
@@ -200,6 +201,23 @@ func UpdateApiGatewayStatus(ctx context.Context, k8sClient client.Client, apiGat
 			return getErr
 		}
 
+		if newStatus.State != operatorv1alpha1.Processing {
+			for i := range newStatus.Conditions {
```
Ressetkk: If you call `newStatus.Conditions` multiple times, initialize it at the beginning of function for better code readability.

Ressetkk (changes requested):
It's OK overall, but there are small nits.

Ressetkk (dismissed):

Ressetkk (approved):

nataliasitko (approved):

Issue #2852, opened 2026-09-11
Tests are executed even though the SSL cert used by Kyma gateway is not created yet


Ressetkk:
Probably bug in api gateway controller / kyma gateway recon. 

Detected here: https://github.com/kyma-project/api-gateway/actions/runs/14289218623/attempts/1

Tests wait for API Gateway CR status that doesn't reflect Kyma gateway status properly. It doesn't take Certificate status into consideration.
As a result tests are being executed on not fully configured system.

We should either:
- observe certificate or kyma-gateway-certs secret status if enableKymaGateway is set
- or at least work it around in tests, so they are paused until kyma-gateway-certs secret is created

See:
![Image](https://github.com/user-attachments/assets/9dcea8b0-ef4b-4994-b05e-e3136bdd5018)

PR #2856, opened 2026-09-14
feat: add wait for kyma gateway certificate CR

Ressetkk:
When we reconcile Gardener CR we don't wait for it to be Ready. This PR adds a requeue until cert is ready by this time the APIGateway CR is in Processing state until it's Ready.

If the certificate flips back to NotReady, we don't change previously Ready AGW CR to "Processing". Though I think we should work on "Ready" and "NotReady" instead...

#2852


werdes72 (dismissed):

nataliasitko (dismissed):

mluk-sap:
The current implementation is stuck in 'Processing' when the Gardener certificate has a final 'Error' status.
It can be reproduced by breaking kyma-tls-cert right after creation:
image

nataliasitko (approved):

mluk-sap (approved):

